Issue #14, opened 2025-07-14
feat: pass in rules and configurations


Boshen:
Not sure about the best approach yet, but oxlint will spawn tsgolint, and pass in rules to run and their configurations.

nrayburn-tech:
Maybe something like this?

```json
[
  {
    "files": ["paths to lint"],
    "config": {
      "rules": {

      }
    },
    "tsconfig": "tsconfig path or maybe the parsed config json"
  }
]
```

There should be 1 object supplied per tsconfig/oxlint config combination.

Files should be all files within the tsconfig that should be linted, meaning ignore paths has already been applied and those files removed.

Config should be similar to the Oxlint config file after overrides have been applied. Rules is the only config option that comes to mind at the moment, but other options could be passed through as well.

Tsconfig is the path to the tsconfig file to use for linting.

Issue #8, opened 2025-07-14
feat: output diagnostics as structured data


Boshen:
Not sure about the best approach yet, but oxlint will spawn tsgolint, and print out the diagnostics data emitted by tsgolint

nrayburn-tech:
I would guess that outputting diagnostics as JSON is probably good enough for a beginning implementation. Maybe something similar to what `oxlint —format=json` looks like.

Is the intention to do all communication through stdin/stdout? Maybe this is fine for now and a new issue can be created in the future if it isn’t sufficient.

PR #21, opened 2025-07-15
Prototype headless mode

### LINKED ISSUE #7: feat: cli accepting paths to lint
Not sure about the best approach yet, but oxlint will spawn tsgolint and pass in a list of paths to lint.

auvred:
Fixes #7, #8, #14

This PR adds a new `tsgolint headless` subcommand. It takes two required arguments: `--tsconfig`, `--cwd`.

Config should be written to the STDIN of the tsgolint process in the JSON format:

```json
{
  "files": [
    {
      "file_path": "/absolute/path/to/file.ts",
      "rules": ["rule-1", "another-rule"]
    }
  ]
}
```

tsgolint will write messages back to the calling party via STDOUT. The format of the messages is as follows:

```
| Payload Size (uint32 LE) - 4 bytes | Message Type (uint8) - 1 byte | Payload |
```

Currently there are two types of message:

- Error - type:`0` ([payload structure](https://github.com/oxc-project/tsgolint/blob/bb0eea711f8804115a7286c7cdcc7916884766c8/cmd/tsgolint/headless.go#L81-L83))
- Diagnostic - type:`1` ([payload structure](https://github.com/oxc-project/tsgolint/blob/bb0eea711f8804115a7286c7cdcc7916884766c8/cmd/tsgolint/headless.go#L33-L72))

Diagnostic messages are issued immediately after the linter has reported them. Therefore they can be read from STDOUT and deserialized while the linter is still running.

---

Example:

```shell
echo "{\"files\":[{\"file_path\":\"$(pwd)/file.ts\",\"rules\":[\"no-unsafe-member-access\",\"require-await\"]}]}" | ./tsgolint headless --tsconfig $(pwd)/tsconfig.json --cwd $(pwd)
```

One of the reported diagnostics:

```json
{
  "range": {
    "pos": 1475,
    "end": 1496
  },
  "rule": "no-unsafe-member-access",
  "message": {
    "id": "unsafeMemberExpression",
    "description": "Unsafe member access .queryResultCacheTable on an `any` value."
  },
  "fixes": [],
  "suggestions": [],
  "file_path": "/<omitted>/tsgolint/benchmarks/typeorm/test/functional/cache/custom-cache-provider.ts"
}

```

### DIFF
--- a/cmd/tsgolint/headless.go
+++ b/cmd/tsgolint/headless.go
@@ -23,16 +23,16 @@ import (
 )
 
 type headlessConfigForFile struct {
-	FilePath string
-	Rules    []string
+	FilePath string   `json:"file_path"`
+	Rules    []string `json:"rules"`
 }
 type headlessConfig struct {
-	Files []headlessConfigForFile
+	Files []headlessConfigForFile `json:"files"`
 }
 
 type headlessRange struct {
-	Pos int
-	End int
+	Pos int `json:"pos"`
+	End int `json:"end"`
 }
 
 func headlessRangeFromRange(r core.TextRange) headlessRange {
@@ -43,8 +43,8 @@ func headlessRangeFromRange(r core.TextRange) headlessRange {
 }
 
 type headlessRuleMessage struct {
-	Id          string
-	Description string
+	Id          string `json:"id"`
+	Description string `json:"description"`
 }
 
 func headlessRuleMessageFromRuleMessage(msg rule.RuleMessage) headlessRuleMessage {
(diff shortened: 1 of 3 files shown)


Boshen:
I assume these are utf16 span offsets?

```
  "range": {
    "pos": 1475,
    "end": 1496
  }
```

Boshen:
This payload can become huge.

```
{
  "files": [
    {
      "file_path": "/absolute/path/to/file.ts",
      "rules": ["rule-1", "another-rule"]
    }
  ]
}
```

In oxlint we can get over 10k files, that's 10k * 10 to 20 rules.

auvred:
> I assume these are utf16 span offsets?

These are UTF-8 offsets. tsgo relies on UTF-8 internally.

auvred:
> In oxlint we can get over 10k files, that's 10k * 10 to 20 rules.

We can group by rules then. Something like:

```json
{
  "rule-1": {
    "files": ["file1.ts"],
  }
}
```

But it might be worth leaving it as it is. I don't think it would take any significant amount of time compared to type-aware linting.

Boshen (approved):

OBITOONDEADO1MZ (approved):

OBITOONDEADO1MZ (approved):

OBITOONDEADO1MZ (approved):

PR #22, opened 2025-07-15
fix(prefer-return-this-type): check class expressions return `this`

43081j:
Class expressions are not currently visited but can encounter this
situation:

```ts
const Foo = class FooClass {
  fn(): FooClass {
    return this;
  }
}
```

This change makes the rule visit such expressions and report the
`FooClass` usage.

If there's a cleaner way to do the conditional stuff around declarations vs
expressions, let me know. In go, afaict, it is inevitable that we have to
have two branches to check the two types.

### DIFF
--- a/internal/rules/prefer_return_this_type/prefer_return_this_type.go
+++ b/internal/rules/prefer_return_this_type/prefer_return_this_type.go
@@ -36,13 +36,18 @@ var PreferReturnThisTypeRule = rule.Rule{
 			return nil
 		}
 
-		checkFunction := func(fn *ast.Node, originalClass *ast.ClassDeclaration) {
+		checkFunction := func(fn *ast.Node, originalClass *ast.Node, originalClassName *ast.DeclarationName) {
 			returnType := fn.Type()
 			body := fn.Body()
 			if returnType == nil || body == nil {
 				return
 			}
-			className := originalClass.Name().Text()
+
+			if originalClassName == nil {
+				return
+			}
+
+			className := originalClassName.Text()
 
 			node := tryGetNameInTypeNode(className, returnType)
 			if node == nil {
@@ -57,7 +62,7 @@ var PreferReturnThisTypeRule = rule.Rule{
 				}
 			}
 
-			classType := ctx.TypeChecker.GetTypeAtLocation(originalClass.AsNode()).AsInterfaceType()
+			classType := ctx.TypeChecker.GetTypeAtLocation(originalClass).AsInterfaceType()
 
 			if ast.IsBlock(body) {
 				hasReturnThis := false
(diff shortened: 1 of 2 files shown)


internal/rules/prefer_return_this_type/prefer_return_this_type.go
```diff
@@ -36,13 +36,18 @@ var PreferReturnThisTypeRule = rule.Rule{
 			return nil
 		}
 
-		checkFunction := func(fn *ast.Node, originalClass *ast.ClassDeclaration) {
+		checkFunction := func(fn *ast.Node, originalClass *ast.Node, originalClassName *ast.DeclarationName) {
```
auvred:
```suggestion
		checkFunction := func(fn *ast.Node, originalClass *ast.Node) {
```

internal/rules/prefer_return_this_type/prefer_return_this_type.go
```diff
@@ -36,13 +36,18 @@ var PreferReturnThisTypeRule = rule.Rule{
...
 			}
-			className := originalClass.Name().Text()
+
+			if originalClassName == nil {
+				return
+			}
+
+			className := originalClassName.Text()
 
 			node := tryGetNameInTypeNode(className, returnType)
```
auvred:
```suggestion
			className := originalClass.Name()
			if className == nil {
				return
			}

			node := tryGetNameInTypeNode(className.Text(), returnType)
```

internal/rules/prefer_return_this_type/prefer_return_this_type.go
```diff
@@ -109,14 +114,16 @@ var PreferReturnThisTypeRule = rule.Rule{
 				}
 
 				if ast.IsFunctionExpression(property.Initializer) || ast.IsArrowFunction(property.Initializer) {
-					checkFunction(property.Initializer, node.Parent.AsClassDeclaration())
+					classNode := node.Parent.ClassLikeData();
+					checkFunction(property.Initializer, node.Parent, classNode.Name())
```
auvred:
```suggestion
					checkFunction(property.Initializer, node.Parent)
```

internal/rules/prefer_return_this_type/prefer_return_this_type.go
```diff
@@ -109,14 +114,16 @@ var PreferReturnThisTypeRule = rule.Rule{
...
 				}
 			},
 			ast.KindMethodDeclaration: func(node *ast.Node) {
-				if !ast.IsClassDeclaration(node.Parent) {
+				if !ast.IsClassLike(node.Parent) {
 					return
 				}
-				checkFunction(node, node.Parent.AsClassDeclaration())
+				classNode := node.Parent.ClassLikeData();
+				checkFunction(node, node.Parent, classNode.Name())
```
auvred:
```suggestion
				checkFunction(node, node.Parent)
```

auvred (changes requested):
We can just call `originalClass.Name()` instead of `originalClass.ClassLikeData().Name()`, so there is no need to pass `originalClassName`.

43081j:
great! i missed that you can call `Name()` on a node

PR #23, opened 2025-07-15
chore: add deploy CD

### LINKED ISSUE #6: infra: github releases
This repo needs to make a github release with compiled binaries as release assets.

Oxlint will add the compiled binary to its npm packages.

The release needs to include https://github.com/oxc-project/oxc/blob/f717da9ba8449afe5b5e9e6730b45efc8927877b/.github/workflows/release_oxlint.yml#L48-L78

```
- os: windows-latest
  target: x86_64-pc-windows-msvc
  code-target: win32-x64

- os: windows-latest
  target: aarch64-pc-windows-msvc
  code-target: win32-arm64

- os: ubuntu-latest
  target: x86_64-unknown-linux-gnu
  code-target: linux-x64-gnu

- os: ubuntu-latest
  target: aarch64-unknown-linux-gnu
  code-target: linux-arm64-gnu

- os: ubuntu-latest
  target: x86_64-unknown-linux-musl
  code-target: linux-x64-musl

- os: ubuntu-latest
  target: aarch64-unknown-linux-musl
  code-target: linux-arm64-musl

- os: macos-latest
  target: x86_64-apple-darwin
  code-target: darwin-x64

- os: macos-latest
  target: aarch64-apple-darwin
  code-target: darwin-arm64
```

pnodet:
Closes #6

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -45,8 +45,4 @@ jobs:
           project_path: "./cmd/tsgolint"
           binary_name: "tsgolint"
           asset_name: ${{ matrix.asset_name }}
-          build_flags: ${{ matrix.build_flags }}
-          build_tags: ${{ matrix.build_tags }}
-          ldflags: ${{ matrix.ldflags }}
-          pre_command: ${{ matrix.pre_command }}
           extra_files: LICENSE README.md
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -35,13 +35,17 @@ jobs:
           - goos: linux
             goarch: amd64
             asset_name: tsgolint-linux-x64-musl
-            build_flags: "-a -ldflags '-linkmode external -extldflags -static'"
+            build_flags: "-a"
             build_tags: "netgo"
+            ldflags: "-s -w"
+            pre_command: "export CGO_ENABLED=0"
           - goos: linux
             goarch: arm64
             asset_name: tsgolint-linux-arm64-musl
-            build_flags: "-a -ldflags '-linkmode external -extldflags -static'"
+            build_flags: "-a"
             build_tags: "netgo"
+            ldflags: "-s -w"
+            pre_command: "export CGO_ENABLED=0"
           - goos: darwin
             goarch: amd64
             asset_name: tsgolint-darwin-x64
@@ -64,4 +68,6 @@ jobs:
           asset_name: ${{ matrix.asset_name }}
           build_flags: ${{ matrix.build_flags }}
           build_tags: ${{ matrix.build_tags }}
+          ldflags: ${{ matrix.ldflags }}
+          pre_command: ${{ matrix.pre_command }}
           extra_files: LICENSE README.md
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,67 @@
+name: CD
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+
+permissions:
+  contents: write
+  packages: write
+
+jobs:
+  releases-matrix:
+    name: Release Go Binary
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - goos: windows
+            goarch: amd64
+            asset_name: tsgolint-win32-x64
+          - goos: windows
+            goarch: arm64
+            asset_name: tsgolint-win32-arm64
+          - goos: linux
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-gnu
+          - goos: linux
+            goarch: arm64
+            asset_name: tsgolint-linux-arm64-gnu
+          - goos: linux
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-musl
+            build_flags: "-a -ldflags '-linkmode external -extldflags -static'"
+            build_tags: "netgo"
+          - goos: linux
+            goarch: arm64
+            asset_name: tsgolint-linux-arm64-musl
+            build_flags: "-a -ldflags '-linkmode external -extldflags -static'"
+            build_tags: "netgo"
+          - goos: darwin
+            goarch: amd64
+            asset_name: tsgolint-darwin-x64
+          - goos: darwin
+            goarch: arm64
+            asset_name: tsgolint-darwin-arm64
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+          submodules: true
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -0,0 +1,73 @@
...
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-gnu
+          - goos: linux
+            goarch: arm64
+            asset_name: tsgolint-linux-arm64-gnu
+          - goos: linux
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-musl
+            build_flags: "-a"
+            build_tags: "netgo"
```
auvred: tsgolint doesn't resolve DNS names

.github/workflows/cd.yml
```diff
@@ -0,0 +1,73 @@
...
+          - goos: linux
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-gnu
+          - goos: linux
+            goarch: arm64
+            asset_name: tsgolint-linux-arm64-gnu
+          - goos: linux
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-musl
+            build_flags: "-a"
```
auvred: From https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies: 

```
-a
	force rebuilding of packages that are already up-to-date.
```

Why do we need `-a` here?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,73 @@
...
+            build_tags: "netgo"
+            ldflags: "-s -w"
```
auvred: Why is the debug information stripped? And why is it stripped exclusively for musl binaries?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,73 @@
...
+          - goos: darwin
+            goarch: amd64
+            asset_name: tsgolint-darwin-x64
+          - goos: darwin
+            goarch: arm64
+            asset_name: tsgolint-darwin-arm64
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
```
auvred: Why do we need to fetch 2 commits?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,73 @@
...
+          submodules: true
+      - uses: wangyoucao577/go-release-action@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          goos: ${{ matrix.goos }}
+          goarch: ${{ matrix.goarch }}
+          goversion: "https://dl.google.com/go/go1.21.5.linux-amd64.tar.gz"
```
auvred: This won't work. tsgolint requires Go 1.24+.

auvred (changes requested):

.github/workflows/cd.yml
```diff
@@ -0,0 +1,73 @@
...
+            asset_name: tsgolint-win32-arm64
+          - goos: linux
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-gnu
+          - goos: linux
+            goarch: arm64
+            asset_name: tsgolint-linux-arm64-gnu
+          - goos: linux
+            goarch: amd64
+            asset_name: tsgolint-linux-x64-musl
```
auvred: Also: we don't use cgo, so binaries built for `GOOS=linux` will work with both glibc and musl. There is no need to publish the same binary twice.

pnodet:
@auvred This should have been a draft sorry, I was testing the gh action.
I've updated the file to simplify and reflect the changes mentioned.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,58 @@
+name: CD
+
+on:
+  workflow_dispatch:
+  pull_request:
```
auvred:
```suggestion
```

Why is this workflow triggered for pull requests?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,58 @@
...
+  push:
+    branches:
+      - main
```
auvred:
```suggestion
```

I don't think publishing a new release for every commit on `main` is a good idea.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,58 @@
...
+    tags:
+      - "v*"
```
auvred:
```suggestion
```

We don't have any versions and tags yet, so this is not needed for now.

.github/workflows/cd.yml
```diff
@@ -0,0 +1,58 @@
...
+            asset_name: tsgolint-linux-x64
+          - goos: linux
+            goarch: arm64
+            asset_name: tsgolint-linux-arm64
+          - goos: darwin
+            goarch: amd64
+            asset_name: tsgolint-darwin-x64
+          - goos: darwin
+            goarch: arm64
+            asset_name: tsgolint-darwin-arm64
```
auvred, line 34: Can't we do something like this?

```yaml
matrix:
  goos: [windows, linux, darwin]
  goarch: [amd64, arm64]
```

.github/workflows/cd.yml
```diff
@@ -0,0 +1,58 @@
...
+          goos: ${{ matrix.goos }}
+          goarch: ${{ matrix.goarch }}
+          goversion: "https://dl.google.com/go/go1.24.5.linux-amd64.tar.gz"
+          project_path: "./cmd/tsgolint"
+          binary_name: "tsgolint"
+          asset_name: ${{ matrix.asset_name }}
+          build_flags: ${{ matrix.build_flags }}
+          build_tags: ${{ matrix.build_tags }}
+          ldflags: ${{ matrix.ldflags }}
+          pre_command: ${{ matrix.pre_command }}
```
auvred: Matrix does not include any of these.

.github/workflows/cd.yml
pnodet: I believe we can but I'm not sure it will be cleaner since we then need to handle asset_name

```
asset_name: tsgolint-${{ matrix.goos == 'windows' && 'win32' || matrix.goos == 'darwin' && 'darwin' || 'linux' }}-${{ matrix.goarch == 'amd64' && 'x64' || 'arm64' }}
```

tmm1:
Thanks for pushing on this. Is it possible to invoke this workflow and get some binaries up for testing?

auvred:
I'll work on it next week. In the meantime, you can clone the repo and follow the instructions in the README to build it yourself - it's pretty straightforward.

tmm1:
Thanks, I did try it. It is a bit unwieldily without a way to control which checks run where.

Is the intention that oxlintrc will control that?

PR #24, opened 2025-07-15
feat: Add Renovate configuration

outslept:
took some of the opts from https://github.com/Boshen/renovate/blob/main/default.json


.github/renovate.json
```diff
@@ -0,0 +1,48 @@
+{
```
camc314, line 1: I think we can just reference the base like this, and have overrides where needed rather than duplicating the whole thing

https://github.com/oxc-project/oxc/blob/fb3d995828ca4ad44694be9e27df660f432145f6/.github/renovate.json#L0-L1

outslept: sure

Issue #5, opened 2025-07-14
infra: daily typescript-go update PR


Boshen:
Commits like https://github.com/oxc-project/tsgolint/commit/6dd2d1f878fc691db008d1825a38dc0a0a0b6e6d is done manually, better done automatically, so we can jump into fixes quickly.

PR #25, opened 2025-07-15
ci: add daily typescript-go update workflow

no-yan:
closes: #5

### DIFF
--- a/.github/workflows/update-typescript-go.yml
+++ b/.github/workflows/update-typescript-go.yml
@@ -1,0 +1,85 @@
+name: Update TypeScript-Go
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+permissions:
+  pull-requests: write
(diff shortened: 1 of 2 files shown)


no-yan:
Fixed the [error](https://github.com/no-yan/tsgolint/actions/runs/16299540802/job/46030265233#step:5:24).

I ran this workflow in my repository, and the following PR was created:
https://github.com/no-yan/tsgolint/pull/1

auvred:
Thank you for taking the initiative, but I’ve decided to proceed with this pull request instead - https://github.com/oxc-project/tsgolint/pull/28.

- Commits made by default `GITHUB_TOKEN` don't trigger other workflows, we have to use special personal access token
- It's better to reuse existing dependency update mechanism (Renovate)
- The approach of creating a new branch and pull request every day adds a lot of noise. Especially if one day is missed and the PR isn't merged in time - the next day there will already be two open pull requests

PR #43, opened 2025-07-18
test(only-throw-error): add tests for yield/await errors

43081j:
We can drop this `TODO` since we do actually infer types of yields and
awaits correctly.

It seems typescript will already give us the awaited type when we try
retrieve the type of an `AwaitExpression`. So we're already checking
those correctly.

Similarly, a `yield` evaluates to whatever gets passed to `next(value)`.
So if we have a `Generator<T, void, Error>`, we know it will evaluate as
an `Error`. This is already happening too.

So I have removed the `TODO` and added tests to ensure it keeps working.


auvred (approved):
Thanks for sending this in. I think it's worth fixing this in typescript-eslint too.

43081j:
i've opened typescript-eslint/typescript-eslint#11418

PR #248, opened 2025-09-18
chore: log file name for each file

Boshen:
This is useful for debugging which files are being linted, also for finding crash with `GOMAXPROCS=1`


internal/linter/linter.go
```diff
@@ -147,6 +147,7 @@ func RunLinterOnProgram(program *compiler.Program, files []*ast.SourceFile, work
 
 			for w := range workloadQueue {
 				for file := range w.queue {
+					log.Print(file.FileName())
```
camc314: should this be behind an if logLevel == debug check?

PR #356, opened 2025-11-01
feat: report error when failure in creating program

7086cmd:
When `tsgo` fails to parse the `tsconfig`, it displays an error and, if fatal, returns `nil` when creating the program. However, we didn't implement it, leading to a segfault.

```bash
(base) ethangoh@ethanwu mobile % ./tsgolint
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104dba570]

goroutine 1 [running]:
github.com/microsoft/typescript-go/internal/compiler.(*Program).Host(0x140001e0300?)
        /Users/ethangoh/Developer/tsgolint/typescript-go/internal/compiler/program.go:292
main.runMain()
        /Users/ethangoh/Developer/tsgolint/cmd/tsgolint/main.go:412 +0x6f8
main.main()
        /Users/ethangoh/Developer/tsgolint/cmd/tsgolint/main.go:525 +0x1c
```

The `tsgo` reports:

```
(base) ethangoh@ethanwu mobile % pnpm tsgo
tsconfig.json:23:5 - error TS5102: Option 'baseUrl' has been removed. Please remove it from your configuration.
  Use '"paths": {"*": ["./*"]}' instead.

23     "baseUrl": ".",
       ~~~~~~~~~


Found 1 error in tsconfig.json:23
```

I added the logic to handle the TypeScript error, but I'm new to Go and don't have a clear idea of how those modules work. I wanted to implement a diagnostic report similar to the lint result, but I couldn't successfully understand it.

Now it looks like:

```
(base) ethangoh@ethanwu mobile % ./tsgolint
error creating TS program:
  ***/tsconfig.json: Option 'baseUrl' has been removed. Please remove it from your configuration.
```

If there are any problems, I'm happy to fix!

### DIFF
--- a/cmd/tsgolint/main.go
+++ b/cmd/tsgolint/main.go
@@ -404,31 +404,14 @@ func runMain() int {
 		UseCaseSensitiveFileNames: host.FS().UseCaseSensitiveFileNames(),
 	}
 
-	program, diagnostics, err := utils.CreateProgram(singleThreaded, fs, currentDirectory, configFileName, host)
+	program, _, err := utils.CreateProgram(singleThreaded, fs, currentDirectory, configFileName, host)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "error creating TS program: %v", err)
 		return 1
 	}
 
 	if program == nil {
-		// Handle diagnostics from program creation
-		fmt.Fprintf(os.Stderr, "error creating TS program:\n")
-		for _, diag := range diagnostics {
-			if diag.File() != nil {
-				// TODO use diagnostic printer
-				fmt.Fprintf(
-					os.Stderr,
-					"  %s: %s\n",
-					diag.File().FileName(),
-					diag.Message(),
-				)
-			} else {
-				fmt.Fprintf(os.Stderr, "  %s\n", diag.Message())
-			}
-		}
-		if len(diagnostics) == 0 {
-			fmt.Fprintf(os.Stderr, "  unknown error creating program\n")
-		}
+		fmt.Fprintf(os.Stderr, "error creating TS program")
 		return 1
 	}
 


camc314:
thanks for looking at this.

Just as a heads up, the non-headless entry points isn't maintained, and isn't recommended to be used.

oxlint-tsgolint should be called via oxlint

7086cmd:
Yeah, and I understand it. However, when calling the oxlint via --type-aware, It would result to "failed to parse" the "null" value; after removing the wrong field in tsconfig this error disappears, so I assume it's related. Hence, there might be a way to report this (failed to create program) to oxlint side, on the other hamd.

camc314:
> Yeah, and I understand it. However, when calling the oxlint via --type-aware, It would result to "failed to parse" the "null" value; after removing the wrong field in tsconfig this error disappears, so I assume it's related. Hence, there might be a way to report this (failed to create program) to oxlint side, on the other hamd.

can you share your versions of oxlint and tsgolint. it should report a diagnostic if this happens not crash

7086cmd:
oxlint: ^1.24.0
  oxlint-tsgolint: ^0.4.0

It simply failed to parse the tsgolint result (null) — it crashes on the Rust side while serializing.

Its result:

```
Error running tsgolint: "exit status: exit status: 0, error: Failed to parse tsgolint diagnostic payload: invalid type: null, expected path string at line 1 column 179"%
```

camc314:
> oxlint: ^1.24.0 oxlint-tsgolint: ^0.4.0
> 
> It simply failed to parse the tsgolint result (null) — it crashes on the Rust side while serializing.
> 
> Its result:
> 
> ```
> Error running tsgolint: "exit status: exit status: 0, error: Failed to parse tsgolint diagnostic payload: invalid type: null, expected path string at line 1 column 179"%
> ```

Update to 1.25.0, and the issue should be fixed

7086cmd:
Yeah. In `oxlint --type-aware`, it reported this. In `tsgolint`, it got a segfault. If there's no need to fix the bare executing way, closing this is okay. On the other hand, I think it's better to output diagnostic-like items, rather than directly showing an error, because `tsgo` gives the fix instruction?

camc314:
> Yeah. In oxlint --type-aware, it reported this. In tsgolint, it got a segfault. If there's no need to fix the bare executing way, closing this is okay. On the other hand, I think it's better to output diagnostic-like items, rather than directly showing an error, because tsgo gives the fix instruction?


The error in the next release is much better. this was just a first pass. https://x.com/Cameron_C2/status/1985342808916455739

PR #371, opened 2025-11-04
Fix implemented rules counter

Perdolique:
I've counted and found that the implemented rules count is 41 of 59, not 40.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ We welcome contributions! See [CONTRIBUTING.md](./CONTRIBUTING.md) for:
 - How to implement new rules
 - Code style and contribution workflow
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,7 +104,7 @@ We welcome contributions! See [CONTRIBUTING.md](./CONTRIBUTING.md) for:
 - How to implement new rules
 - Code style and contribution workflow
 
-## Implemented Rules (40/59)
+## Implemented Rules (41/59)
```
camc314, line 107: i think this should be 42 now.

can you check off no-deprecated as well please

camc314: fixed by #371

PR #389, opened 2025-11-10
feat: implment `no-unnecessary-condition` rules

### LINKED ISSUE #188: [no-unnecessary-condition](https://typescript-eslint.io/rules/no-unnecessary-condition)

sunrabbit123:
resolve #188 

293/293 (100%)

### DIFF
--- a/internal/rules/no_unnecessary_condition/no_unnecessary_condition_test.go
+++ b/internal/rules/no_unnecessary_condition/no_unnecessary_condition_test.go
@@ -2320,7 +2320,7 @@ declare const foo: { bar: number };
 foo.bar ??= 1;
       `,
 			TSConfig: "tsconfig.exactOptionalPropertyTypes.json",
-			Errors:  []rule_tester.InvalidTestCaseError{{MessageId: "neverNullish"}},
+			Errors:   []rule_tester.InvalidTestCaseError{{MessageId: "neverNullish"}},
 		},
 
 		// Function return values with optional chaining
@@ -2569,7 +2569,7 @@ if (test[0]?.a) {
 }
       `,
 			TSConfig: "tsconfig.noUncheckedIndexedAccess.json",
-			Errors:  []rule_tester.InvalidTestCaseError{{MessageId: "neverOptionalChain"}},
+			Errors:   []rule_tester.InvalidTestCaseError{{MessageId: "neverOptionalChain"}},
 		},
 		{
 			Code: `
@@ -2591,7 +2591,7 @@ if (arr[0]) {
 }
       `,
 			TSConfig: "tsconfig.noUncheckedIndexedAccess.json",
-			Errors:  []rule_tester.InvalidTestCaseError{{MessageId: "neverNullish"}},
+			Errors:   []rule_tester.InvalidTestCaseError{{MessageId: "neverNullish"}},
 		},
 		{
 			Code: `
@@ -2601,7 +2601,7 @@ if (arr[42] && arr[42]) {
 }
       `,
 			TSConfig: "tsconfig.noUncheckedIndexedAccess.json",
-			Errors:  []rule_tester.InvalidTestCaseError{{MessageId: "alwaysTruthy"}},
+			Errors:   []rule_tester.InvalidTestCaseError{{MessageId: "alwaysTruthy"}},
 		},
 	})
 }


sunrabbit123:
256/293 (87.4%)

sunrabbit123:
Current Status: 263/294 tests passing (89.5%)

sunrabbit123:
# Known Limitations in `no-unnecessary-condition` Rule

This document explains the technical limitations that prevent two test cases from passing.

## Test Case: `valid-139` - Union of Functions Return Type

### Code
```typescript
type FooUndef = () => undefined;
type FooNum = () => number;
type Foo = FooUndef | FooNum | null;
declare const foo: Foo;
foo?.()?.toExponential();  // Second ?. should be allowed
```

### Expected Behavior
The second `?.` should NOT be flagged as unnecessary because:
- `foo?.()` calls either `FooUndef` (returns `undefined`) or `FooNum` (returns `number`)
- The result type is `undefined | number`
- Since the result can be `undefined`, the second `?.toExponential()` is necessary

### Why It Fails

**Root Cause**: TypeScript Go API's `GetCallSignatures()` does not properly resolve type aliases.

When checking if a union of functions returns nullish:
1. We get the function type: `FooUndef | FooNum | null`
2. Remove nullish parts: `FooUndef | FooNum`
3. Iterate through union parts and call `GetCallSignatures(part)`
4. **Problem**: `GetCallSignatures(FooUndef)` returns an empty array because `FooUndef` is a type alias, not a direct function type

The TypeScript compiler internally resolves type aliases before getting call signatures, but the Go API's `GetCallSignatures()` appears to operate on the alias itself without resolution.

### Attempted Solutions
1. Directly using `GetTypeAtLocation` on the call expression - returns combined type but loses function-specific information
2. Iterating union parts and getting signatures - fails due to type alias issue
3. Using `GetReturnTypeOfSignature` - requires valid signatures which we can't obtain

### Potential Fix
Would require either:
- A TypeScript Go API method to resolve type aliases before querying signatures
- Access to the underlying function type from a type alias
- A `GetResolvedCallSignatures` method that handles aliases

---

## Test Case: `invalid-130` - Required Mapped Type with Template Literal Key

### Code
```typescript
type A = {
  [name in Lowercase<string>]?: {       // OPTIONAL mapped type
    [name in Lowercase<string>]: {      // REQUIRED mapped type
      a: 1;
    };
  };
};

declare const a: A;
a.a?.a?.a;  // Second ?. should be flagged as unnecessary
```

### Expected Behavior
The second `?.` SHOULD be flagged as unnecessary because:
- `a.a` has type `Middle | undefined` (outer mapped type is optional)
- `a.a?.a` accesses property `a` on `Middle` which is a **required** mapped type
- The result is `Inner` (NOT `Inner | undefined`) because the inner mapped type has no `?`
- Therefore, the second `?.a` is unnecessary

### Why It Fails

**Root Cause**: `GetStringIndexType()` returns `nil` for template literal mapped types like `Lowercase<string>`.

The detection flow:
1. Get non-nullish base type: `Middle` = `{ [name in Lowercase<string>]: Inner }`
2. Try `getPropertyOfType(Middle, "a")` → returns `nil` (no explicit property named "a")
3. Try `GetStringIndexType(Middle)` → returns `nil` (template literal types are not regular string index signatures)
4. Fall back to `GetTypeAtLocation(expression)` which returns `Inner | undefined` (includes undefined from optional chain)
5. Since result includes undefined, we don't flag it as unnecessary

### The Core Problem

We cannot distinguish between:
1. **Required mapped type**: `{ [K in Lowercase<string>]: V }` → accessing gives `V`
2. **Optional mapped type**: `{ [K in Lowercase<string>]?: V }` → accessing gives `V | undefined`

Both have `GetStringIndexType()` returning `nil` because `Lowercase<string>` is an intrinsic/template literal type, not a regular string.

For regular string index signatures like `{ [key: string]: V }`:
- `GetStringIndexType()` returns `V`
- We can properly detect the value type

### Attempted Solutions
1. Return `stringIndexType` directly when available - breaks `valid-114`/`valid-115` which need undefined for regular index signatures
2. Strip undefined from `GetTypeAtLocation` result - incorrectly strips undefined for optional mapped types (`valid-155`)
3. Check `GetApparentProperties()` - returns empty for mapped types
4. Check type flags on result - can't distinguish source of undefined (optional chain vs optional mapped type)

### Potential Fix
Would require either:
- A TypeScript Go API method to get index info with optionality flag
- Access to `IndexInfo.isOptional` or similar
- A way to query mapped type modifiers (`?`, `readonly`, etc.)

---

## Summary

| Test | Issue | API Limitation |
|------|-------|----------------|
| valid-139 | Union of functions | `GetCallSignatures()` doesn't resolve type aliases |
| invalid-130 | Template literal mapped type | `GetStringIndexType()` returns nil for `Lowercase<string>` |

Both issues stem from the TypeScript Go API not exposing certain internal type resolution mechanisms that the original TypeScript compiler uses.

jokull:
I got the tests to pass by hammering with Claude Opus - but I don't have the go experience to judge the effort - if anyone's interested though:

I picked up this work on my fork at https://github.com/jokull/tsgolint/pull/2 and addressed the two known limitations you documented:

**valid-139** (Union of Functions): While GetCallSignatures() may not resolve type aliases directly, the implementation now falls back to GetTypeAtLocation() on the call expression, which correctly returns the resolved return type (undefined | number). Since this contains undefined, the optional chain is correctly allowed.

**invalid-130** (Template Literal Mapped Type): Instead of relying on GetStringIndexType() (which returns nil for template literals), I exposed GetMappedTypeModifiers via go:linkname to directly query whether a mapped type has the ? optional modifier. This correctly distinguishes between { [K in Lowercase<string>]?: T } and { [K in Lowercase<string>]: T } regardless of the key type.

sunrabbit123:
> I got the tests to pass by hammering with Claude Opus - but I don't have the go experience to judge the effort - if anyone's interested though:
> 
> I picked up this work on my fork at [jokull#2](https://github.com/jokull/tsgolint/pull/2) and addressed the two known limitations you documented:
> 
> **valid-139** (Union of Functions): While GetCallSignatures() may not resolve type aliases directly, the implementation now falls back to GetTypeAtLocation() on the call expression, which correctly returns the resolved return type (undefined | number). Since this contains undefined, the optional chain is correctly allowed.
> 
> **invalid-130** (Template Literal Mapped Type): Instead of relying on GetStringIndexType() (which returns nil for template literals), I exposed GetMappedTypeModifiers via go:linkname to directly query whether a mapped type has the ? optional modifier. This correctly distinguishes between { [K in Lowercase]?: T } and { [K in Lowercase]: T } regardless of the key type.

cool! 👍 
but, well.. i have some questions

1. Is there a reason you didn’t open the PR yourself?
2. If you don’t plan to open it yourself, would it be okay if I add that commit to this PR?

If you plan to open the PR yourself, I’m also willing to close this PR.

jokull:
Please take over the code :)

sunrabbit123:
@jokull thank you for your contribution 
the draft is converted PR

jokull:
Maybe update the PR description so progress and state of the PR is clear?

sunrabbit123:
ah good idea,

ColemanDunn:
LFG! One of the most meaningful type aware rules IMO

camc314:
Amazing work - Thank you!

Sorry it took me so long to review!

PR #394, opened 2025-11-11
chore: add test util for to get failed count

sunrabbit123:
if we failed test code, we need success rate for report


sunrabbit123:
yeah!, but we get test code from eslint test set, so we(or I) need statistics on how well the test code follows them.
Of course, this is only for development.
Im talking about how much progress we have left.

스크린샷 2025-11-12 오전 11 15 53

Nothing changes except the line with the emoji at the end.

auvred:
Why not just https://github.com/gotestyourself/gotestsum?

sunrabbit123:
good idea, but IMO it's depdendency
I try not to increase my dependence as much as possible.
This is just a personal disposition.

Issue #425, opened 2025-11-15
Remove support for some features that are deprecated in `typescript-eslint`


camchenry:
This a tracking issue to record some thoughts around removing support for some features that are already discouraged or deprecated in `typescript-eslint`. Then, if we do decide to move forward with this, we can refer back to this issue as documentation of that decision.

## Proposal

Remove support for a number of deprecated features within `typescript-eslint`:

- Remove `allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing` in rules:
  - `no-unnecessary-boolean-literal-compare`
  - `strict-boolean-expressions`
- Remove support for [universal string specifiers in `TypeOrValueSpecifier`](https://typescript-eslint.io/packages/type-utils/type-or-value-specifier/#universal-string-specifiers)
  - `allowForKnownSafeCalls` in `no-floating-promises`
  - `allowForKnownSafePromises` in `no-floating-promises`
  - `allow` in `only-throw-error`

#### `allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing`

For example, the option `allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing` is supported in a few lint rules, but there are only a few hundred files across the entirety of indexed code on GitHub that actually configure it:

https://github.com/search?q=%2FallowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing.%2B%28%28true%29%7C%28false%29%29%2F&type=code&p=2

Although marginal, these options do have some cost during deserialization checking if they're enabled, and take up some small amount of memory. So removing these sorts of options does have a very small performance benefit.

#### `allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing`

Rules that support the `allowRuleToRunWithoutStrictNullChecksIKnowWhatIAmDoing` option require the `strictNullChecks` option to be enabled. The `strictNullChecks` option represents a widely used best practice for TypeScript and should be enabled for the safest TypeScript code. In fact, [TypeScript 6.0 will enable `--strict` by default](https://github.com/microsoft/TypeScript/issues/62333) because it is so widely used and highly recommended.

### TypeScript 6.0 and 7.0, and `tsgolint` represent a huge leap forward

As part of the migration to TypeScript 7.0 (`typescript-go`), TypeScript 6.0 will include a number of breaking changes as well:

- Enabling `--strict` by default: https://github.com/microsoft/TypeScript/issues/62333 
[...]

PR #252, opened 2025-09-18
feat: initialize rule generation tool

camc314:
# Add Rule Generator Tool

This PR adds a TypeScript-based rule generator tool to help create new type-aware ESLint rules for TSGolint. The tool:

- Downloads test files from typescript-eslint repository
- Parses test cases to extract valid/invalid examples
- Generates Go rule implementation stubs and test files
- Verifies that rules require type checking (since TSGolint focuses on type-aware rules)

The PR also:
- Adds a new `new-rule` command to the justfile for easy rule generation
- Adds TypeScript type checking for the rulegen tool in CI
- Includes necessary dependencies and configuration files

Example usage: `just new-rule await-thenable`

part of #250

PR #484, opened 2025-12-03
feat: prefer-optional-chaining

### LINKED ISSUE #199: [prefer-optional-chain](https://typescript-eslint.io/rules/prefer-optional-chain)

kalvenschraut:
disclaimer this PR's code was generated entirely using opencode and claude opus 4.5. It has taken a lot of prompting on my end to get it into this state which I believe is a working one. I am not entirely sure how I would be able to test this on my ts codebase. I assume something with pnpm link would be possible?

I had it try look into reasons why my tests were so much slower and it found https://github.com/oxc-project/tsgolint/commit/1b11bfcd7f17ce2d75d2f57954f164944d897762 to be one of the main reasons since the host in this repo has caching while ts go doesnt to my understanding.

I plan to review this over next week and test it out before I mark this as fully ready for external review but want to get some initial feedback if this was even the right direction / major glaring issues with the AI code.

Pretty excited about the speed increases from this as it is a major pain point in my company's code base and this rule and at least one other prefer nullish coalescing would be ones I wanted added asap so decided to try do it with opencode in my free time 

closes #199


internal/rules/prefer_optional_chain/schema.json
```diff
@@ -0,0 +1,42 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "definitions": {
+    "prefer_optional_chain_options": {
+      "type": "object",
+      "properties": {
+        "allowPotentiallyUnsafeFixesThatModifyTheReturnTypeIKnowWhatImDoing": {
```
connorshea, line 7: Probably should just skip adding this option, in light of https://github.com/oxc-project/tsgolint/issues/425

connorshea: Technically this isn't a deprecated config option so there's maybe an argument for keeping it, but 🤷‍♂️

kalvenschraut: yea I guess if its not deprecated then I guess should keep otherwise people who migrate may be confused where it is at IMO

internal/utils/create_program.go
```diff
@@ -16,7 +16,7 @@ import (
 
 func CreateCompilerHost(cwd string, fs vfs.FS) compiler.CompilerHost {
 	defaultLibraryPath := bundled.LibPath()
-	return compiler.NewCompilerHost(cwd, fs, defaultLibraryPath, nil, nil)
+	return NewCompilerHost(cwd, fs, defaultLibraryPath, nil, nil)
```
kalvenschraut, line 19: Here was the change that helped speed up my test cases a lot IDK if this has any side effects other than AI seeing that typescript go didnt do any caching while our own host did

internal/rules/prefer_optional_chain/schema.json
camc314: lets keep it for now, and remove it all at once before we do a stable release

kalvenschraut:
@camc314  was trying to see how you auto generated the tests with my previous PR and I used the script in #252 and it seems to have found issues. I had to modify it a bit to get all the tests outside of the base case ones to pull in. Working through those issues now

internal/utils/create_program.go
kalvenschraut: Seems like it was merged in #522

kalvenschraut:
I think I am going call this good finally. The actual implementation logic I read through but a lot going on so definitely relying on all the tests that were auto brought over by the script and the base case tests I did review to try make sure they match what typescript-eslint tests are trying cover.  With all those tests passing, mainly was looking for places that had unneeded code and found as much as I could.

PR #496, opened 2025-12-09
fix(no-unnecessary-type-assertion): avoid duplicate spaces in autofixer

maschwenk:
It seems like the test expectations here were asserting what looks like something wrong to me, so it's perhaps a misunderstanding on my side

### DIFF
--- a/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
+++ b/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
@@ -1008,7 +1008,8 @@ const foo = (  3 + 5
       `,
 			Output: []string{`
 const foo = (  3 + 5
-  ) /*as*/;
+  ) /*as*/ //as
+  ;
       `,
 			},
 			Errors: []rule_tester.InvalidTestCaseError{
--- a/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
+++ b/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
@@ -211,23 +211,62 @@ var NoUnnecessaryTypeAssertionRule = rule.Rule{
 			if node.Kind == ast.KindAsExpression {
 				ctx.ReportNodeWithFixes(node, msg, func() []rule.RuleFix {
 					s := scanner.GetScannerForSourceFile(ctx.SourceFile, expression.End())
-					asKeywordStart := s.TokenStart()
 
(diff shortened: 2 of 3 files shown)


internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
```diff
@@ -406,7 +406,7 @@ foo!;
 	}, []rule_tester.InvalidTestCase{
 		{
 			Code:   "const a = `a` as const;",
-			Output: []string{"const a = `a`  ;"},
```
maschwenk, line 409: I guess unsure if we're trying to avoid handling formatting, so maybe this is as intended

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
```diff
@@ -1057,7 +1056,7 @@ const foo = 3 + 5/*as*/  ;
 const foo = 3 + 5/*a*/ /*b*/ as number;
```
maschwenk, line 1057: ahh ok, maybe this is the issue. like this is technically stripping comments

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
```diff
@@ -1008,8 +1008,7 @@ const foo = (  3 + 5
       `,
 			Output: []string{`
 const foo = (  3 + 5
-  ) /*as*/  //as
-  ;
```
camc314, line 1012: hmm this is technically a regression

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
camc314: yeah typically my approach with these fixer has been that we shouldn't really care about the formatted output since users will be running a formatter over it after. But if we can trivially update it to produce better output, we should

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
camc314: If we consider:
```
const foo  =
    initialiser /* comment 1 */
as  /* comment 2 */
typeAssertion  /* comment 3 */
```

We can probably take the following approach:
1. create two "fixes, the first to remove the `as` keyword, the second to remove `typeAssertion`
2. we merge the two fixes together if the text between them is only whitespace
3. we move the start of the two fixes earlier if the text before it is white space

PR #499, opened 2025-12-10
Add .DS_Store to gitignore for us Mac heathens

maschwenk:
^ title


Boshen:
Sorry for the close but this should be in your git global ignore 😁

maschwenk:
> Sorry for the close but this should be in your git global ignore 😁

lol fair. will do. realizing this was something i forgot to bring over from my last machine eek

PR #529, opened 2025-12-18
fix: add mutex to prevent data race in ForEachCheckerParallel callback

camc314:
ForEachCheckerParallel runs its callback in parallel goroutines, but the callback was appending to flatQueue without synchronization. This data race could corrupt the slice, resulting in zero-valued checkerWorkload structs where the queue channel is nil.

When workers later tried to receive from the nil queue channel (line 234: `for file := range w.queue`), they would block forever, causing a deadlock:

    fatal error: all goroutines are asleep - deadlock!
    goroutine 1286 [chan receive (nil chan)]:

The fix adds a mutex around the append operation to ensure thread-safe access to the flatQueue slice.

Ref https://github.com/oxc-project/oxc/issues/17070

### DIFF
--- a/internal/linter/linter.go
+++ b/internal/linter/linter.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/typescript-eslint/tsgolint/internal/diagnostic"
 	"github.com/typescript-eslint/tsgolint/internal/rule"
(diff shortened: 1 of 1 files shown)


camc314:
* **#529** <a href="https://app.graphite.com/github/pr/oxc-project/tsgolint/529?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/tsgolint/529?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=529).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

jakebailey:
Yeah, `ForEachCheckerParallel` only guarantees that the checkers themselves are exclusively held, not other bits and bobs used inside of it.

Exciting side effect though 😅

PR #547, opened 2025-12-30
fix: defer debug logging to prevent nil pointer access

no-yan:
_Disclosure: I used ChatGPT to improve my writings and English_

---
Fix nil dereference in debug logging.

`utils.CreateProgram` may return a nil program with internal diagnostics.
When debugging is enabled, we log `program.GetSourceFiles()` before checking diagnostics,
causing a nil pointer dereference.
Move the debug log after the internal diagnostics check.

### DIFF
--- a/internal/linter/linter.go
+++ b/internal/linter/linter.go
@@ -69,16 +69,16 @@ func RunLinter(
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


camc314:
Thank you!

PR #555, opened 2026-01-03
perf: Concurrent tsconfig lookup per file

### LINKED ISSUE #554: Parallelize TSConfig resolution per file to improve performance
Currently, `tsgolint` processes TSConfig resolution sequentially [for each file](https://github.com/oxc-project/tsgolint/blob/d433412baaf50471a6eaf8314cbb1d5a165f04b8/cmd/tsgolint/headless.go#L212-L220). While the BFS within the resolution itself is parallelized, it leads to poor CPU utilization as the cache hit rate increases.
This issue proposes changing the parallelization strategy.

> [!NOTE]
> _Disclosure: I used some AI tools: Claude for prototype, Gemini fo improve my English._

## Problem

In large repositories, TSConfig resolution takes more than 10 seconds.

- Current Strategy: Sequential processing per file. Parallelism is only applied inside the BFS search for `tsconfig.json`.
- Bottleneck: Once the cache is warm, the BFS returns almost instantly. This causes the process to  effectively becomes single-threaded, leaving other CPU core idle.

## Trace
The following trace shows poor process utilization (in Elastic/Kibana).
From 55ms onwards, TSConfig resolution continues, but in the latter half, single-threaded discovery persists for about 1.5 seconds. 
Image

## Proposed changes

Instead of iterating through files sequentially, we should process them in parallel using worker pool.

A secondary benefit is that file-specific goroutine creation is eliminated, reducing scheduling overhead.

## Future work

While this change provides an immediate performance boost, even better solutions could be explored in the future, such as:

- Pipelining the linting process to improve overall throughput.
- Inverting the resolution logic: Instead of finding a TSConfig for each file, we could identify files belonging to each TSConfig.

no-yan:
This PR optimizes `tsconfig` resolution by *75%* by improving concurrency , resulting in ~75% faster resolution in large projects. 
This removes a major bottleneck in large codebases; Kibana now completes resolution in ~3 seconds.

| Project | Before | After | 
|--------|--------|--------|
| elastic/kibana | 14.21s | 3.29s | 
| microsoft/vscode | 1.16s | 0.28s |

fix #554

## Benchmark (vs. main branch)

This PR is 40% faster than main at elastic/kibana (excluding node_modules).

<p>

```
❯ sh ../run.sh cmp
Benchmark 1: trunk
  Time (mean ± σ):     19.865 s ±  0.362 s    [User: 39.647 s, System: 10.195 s]
  Range (min … max):   19.479 s … 20.196 s    3 runs

Benchmark 2: baseline
  Time (mean ± σ):     28.448 s ±  0.666 s    [User: 37.079 s, System: 13.019 s]
  Range (min … max):   27.780 s … 29.113 s    3 runs

Summary
  trunk ran
    1.43 ± 0.04 times faster than baseline
```

</p>
</details> 

> [!NOTE]
> _Disclosure: I used some AI tools: Claude Code for prototype and writing tests, Gemini and ChatGPT for reviewing my code and improve my English._


no-yan:
CI failed by Github API rate-limiting. I'll re-run it later.

Info: GitHub API restricts unauthenticated requests to 60 calls per hour.
See: [Rate limits for the REST API](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28&versionId=free-pro-team%40latest&productId=rest#about-primary-rate-limits)

cmd/tsgolint/headless.go
```diff
@@ -207,32 +207,26 @@ func runHeadless(args []string) int {
...
-				workload.Programs[tsconfig] = append(workload.Programs[tsconfig], normalizedFilePath)
-			}
-			fileConfigs[normalizedFilePath] = config.Rules
-			idx++
+	result := tsConfigResolver.FindTsConfigParallel(normalizedFiles)
+	for file, tsconfig := range result {
+		if tsconfig == "" {
+			workload.UnmatchedFiles = append(workload.UnmatchedFiles, file)
+		} else {
+			workload.Programs[tsconfig] = append(workload.Programs[tsconfig], file)
```
camc314, line 234: nice work, can we add back these logs maybe in a final for loop after here?

no-yan: Thanks!　Fixed at https://github.com/oxc-project/tsgolint/pull/555/changes/4a1b2a5620c89ebc55da7dff1f9592216d471fab .

The progress counter `[5/100]` won't be available since we're processing files in parallel now, but the tsconfig resolution result for each file will still be logged.
Please tell me if you want to print the counter.

camc314: yep it's all good to skip the counter - thanks!

internal/utils/find_tsconfig.go
```diff
@@ -195,6 +197,73 @@ func (r *TsConfigResolver) findConfigWithReferences(
...
+				config: result.configFileName,
+				file:   file,
+			}
+			continue
+		}
+
+		out <- ResolutionResult{
+			config: result.configFileName,
+			file:   file,
+		}
```
camc314: these two bits of code look duplicated?
I don't think we need the `result.configFileName != ""` check

cmd/tsgolint/headless.go
```diff
@@ -207,36 +207,38 @@ func runHeadless(args []string) int {
...
-
-	idx := 0
 	for _, config := range payload.Configs {
 		for _, filePath := range config.FilePaths {
-			if logLevel == utils.LogLevelDebug {
-				log.Printf("[%d/%d] Processing file: %s", idx+1, totalFileCount, filePath)
-			}
-
-			normalizedFilePath := tspath.NormalizeSlashes(filePath)
+			fileConfigs[filePath] = config.Rules
```
camc314: I think this still needs to be normalized before we assign it to file configs (filePath)

no-yan:
Testing with aws/aws-cdk revealed failure to resolve tsconfig. Keeping this PR in draft status until the issue is resolved.

camc314:
> Testing with aws/aws-cdk revealed failure to resolve tsconfig. Keeping this PR in draft status until the issue is resolved.

is that a regression?

no-yan:
EDIT: The "bug" was CLI misconfiguration. I should have used `-D typescipt/...` or `-D @typescript-eslint/...`, instead of `-D typescript-eslint/...`.

---
@camc314 
~~After investigation, this bug is **not** a regression of the latest release or this PR.
For now, i've confirmed:~~

- ~~This was introduced before 82b0feef (#356)~~
- ~~Reproduction conditions: Occurs when passing an absolute path to the CLI~~

I will debug the details tomorrow and create an issue for it.

```sh
# Setup
git clone -depth=1 git@github.com:aws/aws-cdk.git

# Bad
pnpm dlx oxlint -A all -D typescript/no-floating-promises --type-aware /Users/noyan/ghq/github.com/aws/aws-cdk
# shows `typescript(tsconfig-error): Invalid tsconfig`
#       `Found 0 warnings and 13 errors. `


# Good
pnpm dlx oxlint -A all -D typescript/no-floating-promises --type-aware ~/ghq/github.com/aws/aws-cdk
# shows `Found 0 warnings and 542 errors.`
```

camc314:
perfect, if you can address this comment

https://github.com/oxc-project/tsgolint/pull/555#discussion_r2667740419

this should be good to merge.

Thanks for your work on this!

cmd/tsgolint/headless.go
no-yan: Thanks!
Combined with the next `for` loop:
https://github.com/oxc-project/tsgolint/pull/555/changes/215cbbb6fb5aa17ca14e95f0a914d3826525110a

no-yan:
@camc314 
Thank you for your review. I've made the corrections you pointed out.

camc314:
forgot to merge this 🫣

jakebailey:
This is something specific to tsgolint, and not something that can be upstreamed, right?

camc314:
> This is something specific to tsgolint, and not something that can be upstreamed, right?

yeah specific to tsgolint

PR #556, opened 2026-01-03
feat: add jsx-no-leaked-render rule

jokull:
## Summary

Adds a new `jsx-no-leaked-render` rule that prevents falsy numeric values from leaking into JSX when using the `&&` operator.

### The Problem

```tsx
// 😱 Renders "0" visibly in the DOM
{count && <Badge>{count}</Badge>}

// 😱 Renders "NaN" visibly in the DOM  
{items.length && <List items={items}/>}
```

### What This Rule Flags

| Type | Flagged | Reason |
|------|---------|--------|
| `number` | ✅ | Could be `0` or `NaN` which render visibly |
| `bigint` | ✅ | Could be `0n` which renders as "0" |
| `any` | ✅ | Could be any of the above |
| `string` | ❌ | Safe in React 18+ (see below) |
| `boolean` | ❌ | `false` is ignored by React |
| `null`/`undefined` | ❌ | Ignored by React |
| `object` | ❌ | Always truthy |

### Safe Patterns

```tsx
{count > 0 && <Badge>{count}</Badge>}
{!!count && <Badge>{count}</Badge>}
{Boolean(count) && <Badge>{count}</Badge>}
{count ? <Badge>{count}</Badge> : null}
```

### Why Strings Are NOT Flagged

We intentionally do not flag `string` types for two reasons:

#### 1. Empty strings are typically intentional

When a developer writes `{title && <Header>{title}</Header>}`, hiding the component when `title` is an empty string is usually the desired behavior. Unlike `0` which is often a valid count that shouldn't hide content, an empty string typically means "no value".

#### 2. React 18+ treats empty strings as null

As of React 18, empty strings are coalesced to `null` in the reconciler, meaning they create no DOM text node. This was changed in [facebook/react#22807](https://github.com/facebook/react/pull/22807) to fix hydration mismatches.

**Before React 18:**
- Empty string `""` → created empty text node → potential issues

**React 18+:**
- Empty string `""` → treated as `null` → no DOM output → safe

This change also benefits React Native, where the previous behavior of rendering empty strings outside `<Text>` components would cause crashes. With React 18+, empty strings are safely ignored.

From the [React reconciler source](https://jser.dev/react/2022/02/04/how-React-handles-empty-values/):
```javascript
if (
  (typeof newChild === "string" && newChild !== "") ||
  typeof newChild === "number"
) {
  // Only non-empty strings and numbers create text nodes
}
```

### Inspiration

Based on [eslint-plugin-jsx-no-leaked-values](https://github.com/gkiely/eslint-plugin-jsx-no-leaked-values) with enhancements:
- Flags any `number` type, not just literal `0` values
- Uses TypeScript type information for accurate detection
- Allows truthy number literals (`1`, `42`, `-1`) 

### Test Coverage

- 39 valid test cases
- 20 invalid test cases

---

### DIFF
--- a/cmd/tsgolint/main.go
+++ b/cmd/tsgolint/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/typescript-eslint/tsgolint/internal/utils"
 
 	"github.com/typescript-eslint/tsgolint/internal/rules/await_thenable"
+	"github.com/typescript-eslint/tsgolint/internal/rules/jsx_no_leaked_render"
 	"github.com/typescript-eslint/tsgolint/internal/rules/no_array_delete"
 	"github.com/typescript-eslint/tsgolint/internal/rules/no_base_to_string"
 	"github.com/typescript-eslint/tsgolint/internal/rules/no_confusing_void_expression"
@@ -125,6 +126,7 @@ func writeMemProfiles(heapOut string, allocsOut string) {
 
 var allRules = []rule.Rule{
 	await_thenable.AwaitThenableRule,
+	jsx_no_leaked_render.JsxNoLeakedRenderRule,
 	no_array_delete.NoArrayDeleteRule,
 	no_base_to_string.NoBaseToStringRule,
 	no_confusing_void_expression.NoConfusingVoidExpressionRule,
--- a/internal/rules/jsx_no_leaked_render/jsx_no_leaked_render.go
+++ b/internal/rules/jsx_no_leaked_render/jsx_no_leaked_render.go
@@ -1,0 +1,96 @@
+package jsx_no_leaked_render
+
+import (
+	"github.com/microsoft/typescript-go/shim/ast"
+	"github.com/microsoft/typescript-go/shim/checker"
+	"github.com/typescript-eslint/tsgolint/internal/rule"
+	"github.com/typescript-eslint/tsgolint/internal/utils"
+)
+
+func buildLeakedRenderMessage() rule.RuleMessage {
+	return rule.RuleMessage{
+		Id:          "noLeakedConditionalRendering",
+		Description: "Potential leaked value that might cause unintentionally rendered values or rendering crashes.",
+	}
+}
+
+// checkLeakyType returns true if the type could leak a falsy value to JSX
+func checkLeakyType(t *checker.Type) bool {
+	// any type is always problematic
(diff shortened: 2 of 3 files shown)


connorshea:
For tsgolint rules - although I'll defer to cam or Boshen on whether this is correct - I believe we generally want to have a larger discussion before we implement anything that isn't coming directly from typescript-eslint right now.

jokull:
Ok that makes sense. I wasn't aware of that, but it does make sense. I will keep the branch and PR around on my [personal fork](http://github.com/jokull/tsgolint/pull/1), this can be closed and I can reopen at a later date if useful. I tested against real world repo and it's immensely useful, for what it's worth.

remorses:
I think it would be nicer to merge new rules PRs by default with the requirement of
1. pre-fixing the rule with `experimental-` or `unstable-` 
2. set them to disabled by default

This way users can try new rule ideas without forking. Given that every rule is independent and does cannot interfere with other rules this sounds like a low risk decision to me. Only users that enable them could be hit by any bugs

I would find this rule very useful myself.

cc @Boshen

connorshea:
> I think it would be nicer to merge new rules PRs by default with the requirement of
> 
> 1. pre-fixing the rule with `experimental-` or `unstable-`
> 2. set them to disabled by default

this would lead to a massive flood of not-thought-through rules for us to maintain, rules conflicting with each other, bug reports to triage, memory usage problems, etc.

(not to say this rule is bad or anything like that, just speaking to what that policy would do in a broader sense)

remorses:
It's also true no original rule was ever merged. I think the current process is too strict

jokull:
What about allowing sidecar tsgo rules or plugins?

remorses:
I created a separate project that let you easily fork on the fly tsgolint: [lintcn](https://github.com/remorses/lintcn). Basically recompiling tsgolint when you add new rules to the project. 

Example of your rule here: https://github.com/remorses/lintcn/tree/main/.lintcn/jsx_no_leaked_render

You can add custom .go files inside `.lintcn` and run `npx lintcn lint`

camc314:
Thanks for working on this.

I'm going to close this one as not planned for now, but we may re-evaluate in future and try get something similar shipped (or expose an API that allows building rules like this via plugins).

PR #630, opened 2026-02-06
perf(test): cache base FS with cachedvfs for node_modules lookup

no-yan:
> [!NOTE]
> _Disclosure: I used Claude for the problem solving and implemetation._

The `no-deprecated` rule traverses `node_modules/@types/node` for program creation.
This change introduces a cached vfs to share fs across test cases.

```
$ hyperfine --warmup 1  -r 3 \
--command-name 'before' 'cd ../tsgolint_original && go test ./internal/rules/no_deprecated -count=1' \
--command-name 'after'  'go test ./internal/rules/no_deprecated -count=1'
Benchmark 1: before
  Time (mean ± σ):      6.236 s ±  0.339 s    [User: 10.069 s, System: 6.996 s]
  Range (min … max):    5.872 s …  6.543 s    3 runs

Benchmark 2: after
  Time (mean ± σ):      3.045 s ±  0.314 s    [User: 7.283 s, System: 1.628 s]
  Range (min … max):    2.749 s …  3.375 s    3 runs

Summary
  after ran
    2.05 ± 0.24 times faster than before
```


camc314 (approved):
ncie!

PR #631, opened 2026-02-06
perf(test): skip @types resolution in tsconfig fixtures

no-yan:
_Disclosure: I used claude to help identify the root cause._

Improved test execution performance by disabling the default `@types` resolution in fixtures.

- Added `"types": []` to `fixtures/tsconfig.json` to stop loading global types.
- Created `fixtures/tsconfig.includeTypes.json` for tests that explicitly require `node` types (e.g. `no_deprecated`).

**Motivation**
By default, TypeScript compiles [all visible `@types` packages](https://www.typescriptlang.org/tsconfig/#types), which causes unnecessary overhead in our test cases.
While [TypeScript 6.0 plans to disable this behavior by default](https://github.com/microsoft/TypeScript/issues/62195), `tsgo` still seems to include them. So I configured it explicitly.

## Bench
2x faster (tested on `no_floating_promises`)

<p>

```
$ hyperfine --warmup 1  -r 3 \
--command-name 'before' 'cd ../tsgolint_original && go test ./internal/rules/no_floating_promises -count=1' \
--command-name 'after'  'go test ./internal/rules/no_floating_promises -count=1'
Benchmark 1: before
  Time (mean ± σ):      3.895 s ±  0.864 s    [User: 8.085 s, System: 5.746 s]
  Range (min … max):    3.346 s …  4.891 s    3 runs

Benchmark 2: after
  Time (mean ± σ):      1.796 s ±  0.090 s    [User: 1.678 s, System: 0.638 s]
  Range (min … max):    1.725 s …  1.897 s    3 runs

Summary
  after ran
    2.17 ± 0.49 times faster than before
```

</p>
</details>

### DIFF
--- a/internal/rules/no_deprecated/no_deprecated_test.go
+++ b/internal/rules/no_deprecated/no_deprecated_test.go
@@ -8,8 +8,10 @@ import (
 )
 
 func TestNoDeprecatedRule(t *testing.T) {
-	rule_tester.RunRuleTester(fixtures.GetRootDir(), "tsconfig.includeTypes.json", t, &NoDeprecatedRule, []rule_tester.ValidTestCase{
 		t.Parallel()
+	rule_tester.RunRuleTester(fixtures.GetRootDir(), "tsconfig.includeTypes.json", t, &NoDeprecatedRule, []rule_tester.ValidTestCase{
+
+
 		{Code: `/** @deprecated */ var a;`},
 		{Code: `/** @deprecated */ var a = 1;`},
 		{Code: `/** @deprecated */ let a;`},


internal/rules/no_deprecated/no_deprecated_test.go
```diff
@@ -8,8 +8,8 @@ import (
 )
 
 func TestNoDeprecatedRule(t *testing.T) {
-	t.Parallel()
-	rule_tester.RunRuleTester(fixtures.GetRootDir(), "tsconfig.json", t, &NoDeprecatedRule, []rule_tester.ValidTestCase{
+	rule_tester.RunRuleTester(fixtures.GetRootDir(), "tsconfig.includeTypes.json", t, &NoDeprecatedRule, []rule_tester.ValidTestCase{
+		t.Parallel()
```
camc314:
```suggestion
		t.Parallel()
	rule_tester.RunRuleTester(fixtures.GetRootDir(), "tsconfig.includeTypes.json", t, &NoDeprecatedRule, []rule_tester.ValidTestCase{

connorshea:
👀 did this change really cut the test-go job from 12m down to 6m?? That's amazing!

EDIT: Actually, maybe the diff was caused by https://github.com/oxc-project/tsgolint/pull/630? It looks like that came first and is the first commit to have run in 6m. Nice work @no-yan!!

jakebailey:
Can you see why we're changing the default to `types: []` in 6.0+?  😄

no-yan:
@connorshea Thank you for kind words!

PR #635, opened 2026-02-07
ci: use `just fmt` in autofix.ci

no-yan:
Ran `just fmt` in autofix.ci.

Generated with Codex.


.github/workflows/autofix.yml
```diff
@@ -25,7 +25,11 @@ jobs:
 
       - uses: oxc-project/setup-node@8958a8e040102244b619c4a94fccb657a44b1c21 # v1.0.6
 
-      - run: pnpm run fmt
+      - uses: extractions/setup-crate@4993624604c307fbca528d28a3c8b60fa5ecc859 # v1.4.0
+        with:
+          repo: casey/just
+
+      - run: just fmt
```
connorshea, line 32: Is there a reason not to use `oxc-project/setup-rust` here?

no-yan: The `extractions/setup-crate` script only downloads pre-built binaries from GitHub Releases.
Therefore, Rust setup is not required here.

camc314: yeah i saw this as well - i'm not sure if `extractions/setup-crate`, is the ideal crate - maybe:
```
- uses: extractions/setup-just@v3
  with:
    just-version: 1.5.0  # optional semver specification, otherwise latest
```

per the just docs?

Issue #650, opened 2026-02-10
Add support for labeled spans in the headless output


camchenry:
For parity with `oxlint`, we should have something akin to [`LabeledSpan`](https://docs.rs/oxc-miette/latest/miette/struct.LabeledSpan.html) that allows us to associate arbitrary ranges in a given file with the diagnostic itself.

This will allow us to enrich diagnostics with much more information than even `typescript-eslint` provides, and give a better experience when doing type-aware linting.

For example, we could add annotations to identifiers with their calculated types when we encounter a type error of some sort.

PR #651, opened 2026-02-10
feat: allow labeled ranges for diagnostics in headless output

camchenry:
- closes https://github.com/oxc-project/tsgolint/issues/650
- depends on https://github.com/oxc-project/oxc/pull/19201

This adds support for labeled ranges being returned by the tsgolint headless mode. This should be a backwards compatible change, as we can simply ignore this additional data on the `oxlint` side. Let's check it, but I think that old versions of `oxlint` won't serialize this field.

## Changes

This adds a new `ReportDiagnostic` that allows completely customizing how a `RuleDiagnostic` is reported. This is like a superset of all the other diagnostic reporting functions. So, this naturally allows setting of the new `LabeledRanges` field as well. Diagnostic functions in rules can then be refactored to return `RuleDiagnostic` instead of `RuleMessage`. This will also be useful in the future in case we add more fields to the diagnostic struct.

## Example

I used the `no-unnecessary-condition` rule diagnostic as an example because it is a good example where we lack contextual information currently, but it makes a lot more sense when we can add contextual labels with the type info.

Example output from running on `test.ts`:

```ts
// test.ts
const left: boolean = true;
const right: string = "foo";

if (left === right) {
  console.log("This will never be logged");
}
```

```json
{
  "kind": 0,
  "message": {
    "id": "noOverlapBooleanExpression",
    "description": "This condition will always return the same value since the types have no overlap."
  },
  "file_path": "/path/to/tsgolint/test.ts",
  "labeled_ranges": [
    { "label": "true", "range": { "pos": 62, "end": 66 } },
    { "label": "string", "range": { "pos": 70, "end": 76 } }
  ],
  "rule": "no-unnecessary-condition"
}
```

### DIFF
--- a/cmd/tsgolint/headless.go
+++ b/cmd/tsgolint/headless.go
@@ -98,11 +98,21 @@ const (
 	headlessDiagnosticKindTsconfig
 )
 
+// A labeled span of source code. Useful for highlighting additional info related to a diagnostic in the context
+// of the source code.
+type headlessLabeledRange struct {
+	// The text label associated with this range.
+	Label string `json:"label"`
+	// The range in the source file that this label applies to.
+	Range headlessRange `json:"range"`
+}
+
 type headlessDiagnostic struct {
-	Kind     headlessDiagnosticKind `json:"kind"`
-	Range    *headlessRange         `json:"range,omitempty"`
-	Message  headlessRuleMessage    `json:"message"`
-	FilePath *string                `json:"file_path"`
+	Kind          headlessDiagnosticKind `json:"kind"`
+	Range         *headlessRange         `json:"range,omitempty"`
+	Message       headlessRuleMessage    `json:"message"`
+	FilePath      *string                `json:"file_path"`
+	LabeledRanges []headlessLabeledRange `json:"labeled_ranges,omitempty"`
 
 	// Only for kind="rule"
 	Rule        *string              `json:"rule,omitempty"`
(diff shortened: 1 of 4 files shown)


camchenry:
* **#651** <a href="https://app.graphite.com/github/pr/oxc-project/tsgolint/651?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/tsgolint/651?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=651).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

camc314:
### Merge activity

* **Feb 11, 9:42 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Feb 11, 9:42 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Feb 11, 9:50 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

Issue #104, opened 2025-08-18
More type-aware rules


Boshen:

brunocroh:
Hello, I would like to contribute and help with this initiative. Do you have at least one example of how to handle these? Cheers.

Boshen:
> Hello, I would like to contribute and help with this initiative. Do you have at least one example of how to handle these? Cheers.

* https://github.com/oxc-project/tsgolint/pull/222

PR #659, opened 2026-02-10
feat: add no-invalid-void-type rule

Limerio:
This commit introduces the `no-invalid-void-type` rule, which disallows the use of `void` as a type in certain contexts, such as union types and generic type parameters. It includes new test cases and example code to validate the rule's functionality. Additionally, updates were made to the end-to-end tests and snapshots to reflect this new rule.


camc314:
Thanks for working on this.

This rule does not need type information to run, hence should be implemented in oxlint, not tsgolint.

https://github.com/oxc-project/oxc/issues/2180

See the list of rules we're porting here: #104

Limerio:
Sorry for that. ^^' Thank you!

camc314:
All good - thanks for looking at this - We'd be very happy to accept a contrbution porting the rule over at https://github.com/oxc-project/oxc !

Limerio:
For sure, I'll do it!

PR #666, opened 2026-02-12
test: add snapshots of test output

camchenry:
Adds snapshots of the diagnostic output. This is immensely helpful as we improve the quality of diagnostics, since we can easily see the effect this will have downstream on `oxlint` right away.

This uses a custom snapshot tool, based on `go-snaps`, which I used for my initial version of this. But I decided that to keep things simple, we only need very little from the library and it would be easier to maintain this small file instead of a whole dependency in our tree.


camchenry:
* **#666** <a href="https://app.graphite.com/github/pr/oxc-project/tsgolint/666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/oxc-project/tsgolint/666?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/oxc-project?ref=merge-queue-instructions-in-stack-comment&prId=666).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

camchenry:
@camc314 can I get your early thoughts on this? for the diagnostic work I'm doing, I thought it would be really helpful to have a preview of the diagnostics that we emit, similar to how we have it set up in `oxlint`.

the main thing I'm unsure about is that I added a dependency on `go-snaps` which is like an `insta`-like library for Go. I'm hesitant to add new dependencies though, but I also don't want to hassle with stuff like Windows support (especially when it comes to paths). If it seems acceptable, I'll get this ready for review. Otherwise, we could consider inling a very minimal version of the library that just emits all of the output to a file for every test run and then errors if there is any `git diff`

camchenry:
@camc314 I've tested this locally and I think this will work fine for us - I feel good enough about this approach that I'm going to move forward with this so that I can keep working on the diagnostics. If you disagree, it is one line change to remove it from the rule tester and we can delete the snapshotting code and snapshots easily. But for now this will make checking the diagnostics so so so much easier than what I have been doing manually with oxlint.

camc314:
Seems reasonable to me - I couldn't initially tell wether this was to prevent regressions, but seeing you want to use it to improve the diagnostic messages it makes a lot of sense.

Although the tester has support for expscting a diagnostic in a particular line/col, it can be touch to work out exactly where they are

PR #673, opened 2026-02-12
perf(patches): cache ReadFile in cachedvfs via patch

no-yan:
Add a patch to cache `ReadFile` results in `cachedvfs`.
`cachedvfs` already caches almost everything, but not `ReadFile`.

I investigated whether the typescript-go team intentionally avoids caching this, but couldn't deduce their intent from the [PRs](https://github.com/microsoft/typescript-go/issues/996/) or commits.

## Output

1.2x faster on aws/aws-cdk.

```
# pnpm dlx oxlint -A all -D typescript/no-floating-promises --type-aware --ignore-pattern '**/*js.snapshot'                                                                 

Benchmark 1: trunk
  Time (mean ± σ):     12.186 s ±  0.623 s    [User: 47.060 s, System: 10.093 s]
  Range (min … max):   11.373 s … 12.820 s    5 runs

Benchmark 2: baseline
  Time (mean ± σ):     15.054 s ±  1.038 s    [User: 54.184 s, System: 18.250 s]
  Range (min … max):   13.928 s … 16.544 s    5 runs

Summary
  trunk ran
    1.24 ± 0.11 times faster than baseline
```

🤖 Generated with [Claude                                                  Code](https://claude.com/claude-code)"


no-yan:
@camc314 @camchenry 
It worked well enough, but is this correct?

jakebailey:
This cached FS was only meant to cache metadata about the FS, as a replacement to very complicated caching in the module resolver. We don't typically read files more than once, since Program is the only one reading files.

It may be that it could be helpful in your case, though?

no-yan:
Thanks so much for the context, Jake! That makes a lot of sense.

In our case, we create multiple Programs (one per tsconfig), so the same files end up being read multiple times across them. That's why caching ReadFile gives us the ~1.2x speedup here.

We'll keep this as a local patch on our side. Really appreciate you taking the time to explain the design intent!

camc314 (approved):
I thought we already did this 🤔 

If you're interested - do you mind taking a look to see if we parse the bound source file anywhere? or whether we re-parse + bind it for every program (even if project have some of the same source files)

camc314:
> 1.2x faster on aws/aws-cdk.

Which files does this really benefit ?  It it node_modules files or actual source files?

no-yan:
@camc314 

> Which files does this really benefit ? It it node_modules files or actual source files?

It benefits both👍

> If you're interested - do you mind taking a look to see if we parse the bound source file anywhere? or whether we re-parse + bind it for every program (even if project have some of the same source files)

Good point, I'll take a look into this tomorrow. Thank you for pointing that out!

Issue #661, opened 2026-02-11
[strict-void-return](https://typescript-eslint.io/rules/strict-void-return/)


connorshea:
This was missing from the original list of rules. https://typescript-eslint.io/rules/strict-void-return/

It was only added relatively recently (a month ago).

PR #676, opened 2026-02-12
feat: implement strict-void-return

Limerio:
Introduced the `strict-void-return` rule to enforce void return types in appropriate contexts.
fixes https://github.com/oxc-project/tsgolint/issues/661

### DIFF
--- a/e2e/__snapshots__/snapshot.test.ts.snap
+++ b/e2e/__snapshots__/snapshot.test.ts.snap
@@ -3320,6 +3320,19 @@ exports[`TSGoLint E2E Snapshot Tests > should generate consistent diagnostics sn
     ],
   },
   {
+    "file_path": "fixtures/basic/rules/strict-void-return/index.ts",
+    "kind": 0,
+    "message": {
+      "description": "Value returned in a context where a void return is expected.",
+      "id": "nonVoidReturn",
+    },
+    "range": {
+      "end": 58,
+      "pos": 54,
+    },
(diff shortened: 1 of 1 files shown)


connorshea:
This PR appears to be missing a lot of tests from the original rule that need to be ported: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/strict-void-return.test.ts

Limerio:
> This PR appears to be missing a lot of tests from the original rule that need to be ported: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/tests/rules/strict-void-return.test.ts

I'll handle that soon!

internal/rules/strict_void_return/strict_void_return_test.go
camc314, line 1: not all test have been ported.

The upstream rule has 210 test cases, this has ~73.

I'm going to mark this as draft so I can keep track of what i need to review, please mark as ready for review once the rest of the tests have been brought across - thanks

Limerio: All tests have been ported.

camc314: Thank you

camc314 (commented):
~I just realized this is missing a test case in e2e snapshot.~

~I'll add one in a follow up PR~

My own tests have come back to bite me - this PR it is!

Issue #574, opened 2026-01-10
crash: using oxlint --type-aware goroutine crash


rubnogueira:
I opened #297 in the past, decided to try it, and it worked, but now when I was integrating it into our codebase, I got the crash:

### How to trigger
❯ oxlint --type-aware

oxlint without --type-aware flag works great.

---
❯ oxlint --version
Version: 1.38.0

❯ tsgo --version
Version 7.0.0-dev.20260109.1

Apparently tsgolint doesn't have the version flag, a suggestion here as well 😄 
tsgolint - 0.11.0

---

❯ which oxlint
/Users/myuser/.local/share/mise/installs/npm-oxlint/1.38.0/bin/oxlint

❯ which tsgolint
/Users/myuser/.local/share/mise/installs/npm-oxlint-tsgolint/0.11.0/bin/tsgolint

❯ which tsgolint
/Users/myuser/.local/share/mise/installs/npm-typescript-native-preview/7.0.0-dev.20260109.1/bin/tsgo

### Trace:

```
2026/01/10 12:51:34.894074 Unmatched files found: ...<hundreds of files>..., /Users/myuser/monorepo/libs/myproject/src/file1.ts, /Users/myuser/monorepo/libs/myproject/src/file2.ts, /Users/myuser/monorepo/libs/myproject/src/file3.ts, /Users/myuser/monorepo/libs/myproject/src/file4.ts, /Users/myuser/monorepo/libs/myproject/src/file5.ts' to be in program '/Users/myuser/monorepo/libs/myproject/tsconfig.json'

goroutine 1 [running]:
github.com/typescript-eslint/tsgolint/internal/linter.RunLinter(0x0, {0x14000024004, 0x4f}, {0x140001fe6c0?, {0x14000e0e280?, 0x14000133ac0?, 0x10333dff8?}}, 0x8, {0x10334a700, 0x140001248e0}, ...)
        /home/runner/work/tsgolint/tsgolint/internal/linter/linter.go:111 +0xcec
main.runHeadless({0x14000132060, 0x0, 0x0})
        /home/runner/work/tsgolint/tsgolint/cmd/tsgolint/headless.go:353 +0xa3c
main.runMain()
        /home/runner/work/tsgolint/tsgolint/cmd/tsgolint/main.go:355 +0x84
main.main()
        /home/runner/work/tsgolint/tsgolint/cmd/tsgolint/main.go:567 +0x1c
```

[...]

PR #686, opened 2026-02-15
fix: disallow circular project references

no-yan:
### Summary
Improves parsing performance by disabling `UseSourceOfProjectReference` and adds proper error handling for circular project references.

### Background

Previously, we enabled `UseSourceOfProjectReference` to fix panics caused by circular project references (#574). However, this option forces the parser to use source files instead of declaration files, which is inherently slow. 

Furthermore, upstream TypeScript and Tsgo [don't support circular project references](https://github.com/microsoft/TypeScript/blob/cdc205d5e6338d96b0fb54657af33d473d480a9c/src/compiler/diagnosticMessages.json#L5435-L5438).
Indeed, running tsgo -b (build mode) on #574's repro produces this error.

Rather than supporting this case, we now detect circular references and emit a clear diagnostic to guide users.

### Benchmark

1.29 times faster on aws-cdk

<p>

```
Benchmark 1: trunk
  Time (mean ± σ):      7.260 s ±  0.252 s    [User: 32.910 s, System: 4.898 s]
  Range (min … max):    6.997 s …  7.500 s    3 runs

Benchmark 2: baseline
  Time (mean ± σ):      9.396 s ±  0.150 s    [User: 41.610 s, System: 7.114 s]
  Range (min … max):    9.243 s …  9.542 s    3 runs

Summary
  trunk ran
    1.29 ± 0.05 times faster than baseline
```

</p>
</details> 

> [!NOTE]
> _Disclosure: I used some AI tools: Codex for implement my plan and gemini for improve my english._

### DIFF
--- a/internal/rule_tester/__snapshots__/only-throw-error.snap
+++ b/internal/rule_tester/__snapshots__/only-throw-error.snap
@@ -349,6 +349,15 @@ Message: Expected an error object to be thrown.
    4 | });
 ---
 
+[TestOnlyThrowErrorRule/invalid-47 - 1]
+Diagnostic 1: object (3:7 - 3:19)
+Message: Expected an error object to be thrown.
+   2 | import { createError } from 'errors-like';
+   3 | throw createError();
+     |       ~~~~~~~~~~~~~
+   4 |       
+---
+
 [TestOnlyThrowErrorRule/invalid-5 - 1]
 Diagnostic 1: object (1:7 - 1:10)
 Message: Expected an error object to be thrown.
--- a/internal/rule_tester/rule_tester.go
+++ b/internal/rule_tester/rule_tester.go
@@ -30,6 +30,7 @@ type ValidTestCase struct {
 	Options  any
 	TSConfig string
 	Tsx      bool
+	Files    map[string]string
 }
 
 type InvalidTestCaseError struct {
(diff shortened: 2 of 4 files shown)


internal/linter/linter.go
```diff
@@ -108,7 +108,15 @@ func RunLinter(
 			}
 			log.Printf("Program source files (%d): %s", len(programFiles), strings.Join(programFiles, ", "))
 
-			panic(fmt.Sprintf("Expected file '%s' to be in program '%s'", unmatchedFilesString, configFileName))
+			for _, f := range unmatchedFiles {
+				onInternalDiagnostic(diagnostic.Internal{
+					Id:          "circular-project-references",
+					Description: fmt.Sprintf("File was not found in program '%s'. This is likely caused by circular project references, which are not supported by TypeScript (TS6202).", configFileName),
```
camc314, line 114: why doesn't the tsconfig parsing/ error matching produce this error code?

no-yan: 1. This error appears only in "build mode" (tsgo -b).
2. Since tsgolint does not use build mode, this error will not occur.


The reason of `1` is likely because references are a no-op in non-"build" mode.
It remains unclear why files are missing from the program created by tsgolint; there can be an issue within find_tsconfig.go.

rubnogueira:
I was notified since you commented in my other post, but I'm not following your PR.

While I understand that when I run with tsc -b may face circular reference errors, would it mean that it would throw an error for such cases with tsgolint?

no-yan:
@rubnogueira 
Yes, exactly. Instead of crashing, I plan to display clear error message.

rubnogueira:
I'm not following. Since eslint/typescript-eslint works perfectly, why the need for tsgolint to display an error message and behave differently?

no-yan:
Thanks for the feedback.

I’m proposing this change because circular references are likely configuration errors.

Project References are option which is used to speed up build.
However, circular dependencies prevent us from running tsgo -b. This means there is no way to achieve the intended build speedup.

If there are valid use cases for composite projects with circular refs that I'm missing, please let me know. I'd love to hear your thoughts!

internal/linter/linter.go
no-yan: I investigated find_tsconfig.go and confirmed it works correctly. It maps files to the right tsconfig even with circular project references.

PR #702, opened 2026-02-20
fix(dot-notation): false positive for non-identifier property names in JSX files

### LINKED ISSUE #701: Bug: dot-notation false positive for property names containing hyphens
### Description

`dot-notation` incorrectly flags bracket notation for property names that contain hyphens (or other characters that make them invalid JS identifiers), such as HTTP header names.

### Reproduction

```ts
const headers = response.headers;
const traceId = headers["X-Amzn-Trace-Id"];
```

**Diagnostic produced:**
```
typescript-eslint(dot-notation): ["X-Amzn-Trace-Id"] is better written in dot notation.
```

### Expected behavior

No diagnostic. `X-Amzn-Trace-Id` contains hyphens and is not a valid JavaScript identifier — `headers.X-Amzn-Trace-Id` would be parsed as `headers.X - Amzn - Trace - Id` (subtraction), not a property access.

### Root cause

Before suggesting dot notation, the rule should verify the property name matches the JS identifier pattern (`/^[a-zA-Z_$][a-zA-Z0-9_$]*$/` or equivalent). Any name containing hyphens, spaces, or other non-identifier characters must use bracket notation and should be exempt.

This is consistent with the behavior of the upstream [`@typescript-eslint/dot-notation`](https://typescript-eslint.io/rules/dot-notation/) rule and the base ESLint `dot-notation` rule.

wagenet:
## Summary

Fixes #701

- In JSX files (`LanguageVariantJSX`), `IsIdentifierText` considers hyphens valid identifier characters (for JSX tag names like `my-component`).
- This caused `dot-notation` to incorrectly flag `headers["X-Amzn-Trace-Id"]` as preferring dot notation in `.tsx` files.
- Fix: use `core.LanguageVariantStandard` when checking whether a property name is a valid JS identifier for dot notation access, since JavaScript property dot-access semantics are always standard regardless of file type.

## Test plan

- [x] Added valid test cases for `a['X-Amzn-Trace-Id']` in both `.ts` and `.tsx` contexts
- [x] Added valid test case for `a['has space']`
- [x] All `dot_notation` tests pass
- [x] All other tests pass (pre-existing `no_deprecated` failures unrelated)

### DIFF
--- a/patches/0001-Parallel-readDirectory-visitor.patch
+++ b/patches/0001-Parallel-readDirectory-visitor.patch
@@ -1,7 +1,7 @@
-From ba1ee831d4ba332e12627a1918b26a19e7e7cf55 Mon Sep 17 00:00:00 2001
+From 7983defbc7e01acc34998bafb44f4685324d54d3 Mon Sep 17 00:00:00 2001
 From: auvred <aauvred@gmail.com>
 Date: Wed, 18 Jun 2025 15:24:34 +0300
-Subject: [PATCH 1/5] Parallel readDirectory visitor
+Subject: [PATCH 1/7] Parallel readDirectory visitor
 
 ---
  internal/vfs/utilities.go | 104 ++++++++++++++++++++++++++++----------
--- a/patches/0002-Adapt-project-service-for-single-run-mode.patch
+++ b/patches/0002-Adapt-project-service-for-single-run-mode.patch
@@ -1,14 +1,14 @@
-From d0efefdc1fc4a9293e0a7b80c32afb673407e0fc Mon Sep 17 00:00:00 2001
+From f513e766e8642da10f651ecc32c47c23e71ee453 Mon Sep 17 00:00:00 2001
 From: Cameron Clark <cameron.clark@hey.com>
 Date: Fri, 14 Nov 2025 11:09:52 +0000
-Subject: [PATCH 2/5] Adapt project service for single run mode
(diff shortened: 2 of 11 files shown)


internal/rules/dot_notation/dot_notation_test.go
```diff
@@ -49,6 +49,10 @@ func TestDotNotation(t *testing.T) {
 		{Code: "\ntype Key = Lowercase<string>;\ntype Foo = {\n  BAR: boolean;\n  [key: Lowercase<string>]: number;\n};\ndeclare const foo: Foo;\nfoo['bar'];\n      ", TSConfig: "tsconfig.noPropertyAccessFromIndexSignature.json"},
 		{Code: "\ntype ExtraKey = `extra${string}`;\n\ntype Foo = {\n  foo: string;\n  [extraKey: ExtraKey]: number;\n};\n\nfunction f<T extends Foo>(x: T) {\n  x['extraKey'];\n}\n      ", TSConfig: "tsconfig.noPropertyAccessFromIndexSignature.json"},
 		{Code: "\ntype Foo = {\n  [key: string]: number;\n};\ndeclare const foo: Foo;\nfoo[`key_baz`];\n      ", TSConfig: "tsconfig.noPropertyAccessFromIndexSignature.json"},
+		// Hyphenated property names are not valid JS identifiers and must use bracket notation
```
camc314:
```suggestion
```

camc314 (approved):
Thank you!

PR #709, opened 2026-02-21
fix(no-deprecated): detect deprecation on inherited methods with no child override

wagenet:
## Summary

In typescript-go, \`Checker_getResolvedSignature\` may not return a declaration for methods inherited from a base class when the child provides no override. The existing signature-based check silently produces no declaration, leaving the deprecation undetected.

- Add \`getJsDocDeprecation(aliasedSymbol)\` as a fallback in the method/function branch of \`getCallLikeDeprecation\`, guarded by \`signatureDecl == nil\`. The guard ensures the fallback never fires for overloaded functions (which always resolve a concrete declaration), preventing false positives on non-deprecated overloads.

## Test plan

- [ ] \`invalid-143\`: \`child.searchPaths()\` where \`Child extends Base {}\` and \`Base.searchPaths\` is \`@deprecated\` — now flagged
- [ ] \`invalid-144\`: same but with deprecation reason string (\`deprecatedWithReason\`)
- [ ] \`valid-22/23/25/26/29/33\`: overloaded functions where only a non-deprecated overload is called — confirmed no regression

### DIFF
--- a/internal/rules/no_deprecated/no_deprecated.go
+++ b/internal/rules/no_deprecated/no_deprecated.go
@@ -232,11 +232,6 @@ var NoDeprecatedRule = rule.Rule{
 				if isDeprecated {
 					return true, reason
 				}
-				// Fallback: if the signature had no declaration (typescript-go limitation for
-				// inherited methods), check the symbol's own declarations directly.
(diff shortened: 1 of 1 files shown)


camc314 (approved):
thank you!

PR #723, opened 2026-02-22
perf(no-unnecessary-type-arguments): use isTypeIdenticalTo + flags checks

kalvenschraut:
AI wrote this when trying to find why a rule was so slow against my company's internal codebase.
Before these changes 
image

after

image

### DIFF
--- a/e2e/__snapshots__/snapshot.test.ts.snap
+++ b/e2e/__snapshots__/snapshot.test.ts.snap
@@ -56,7 +56,7 @@ exports[`TSGoLint E2E Snapshot Tests > should generate consistent diagnostics sn
       "id": "await",
     },
     "range": {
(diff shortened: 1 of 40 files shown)


shim/checker/extra-shim.json
```diff
@@ -31,6 +31,7 @@
       "getBaseTypeOfLiteralType",
       "getWidenedType",
       "isTypeAssignableTo",
+      "isTypeIdenticalTo",
       "isTypeStrictSubtypeOf",
```
connorshea: You may be able to remove this from the shim entirely, it's not used anywhere else.

kalvenschraut: done

no-yan:
Wonderful!

PR #732, opened 2026-02-22
fix(prefer-optional-chaining): various false positives discovered

kalvenschraut:
Trying get this rule going in my companies internal code base now that I have oxlint setup and found what I would consider false positives. Used AI to help come up with fixes and looked over the changes/verified they fix everything in my codebase.

### DIFF
--- a/internal/rules/no_useless_default_assignment/no_useless_default_assignment.go
+++ b/internal/rules/no_useless_default_assignment/no_useless_default_assignment.go
@@ -171,20 +171,6 @@ func getAssignmentTargetNode(node *ast.Node) *ast.Node {
 	return nil
 }
 
-func hasConditionalInitializer(node *ast.Node) bool {
-	if node == nil || node.Parent == nil {
-		return false
-	}
-
-	parent := node.Parent
-	if ast.IsVariableDeclaration(parent) && parent.Initializer() != nil {
-		initializer := ast.SkipParentheses(parent.Initializer())
-		return ast.IsConditionalExpression(initializer) || ast.IsLogicalExpression(initializer)
-	}
-
-	return hasConditionalInitializer(parent)
-}
-
 func hasPropertyInAllBranches(expression *ast.Node, propertyName string) bool {
 	expression = ast.SkipParentheses(expression)
 	if ast.IsObjectLiteralExpression(expression) {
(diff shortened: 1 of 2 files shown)


connorshea:
Have you compared the result to the original typescript-eslint rule? We generally defer to the original rule in terms of what behavior is considered correct.

kalvenschraut:
I had the rule enabled prior when using eslint and it didnt report the error. I did verify this by checking out a branch that still has this rule on for eslint. The oxlint rule actually reported a couple other cases that I considered correct and modified the code to resolve 🤷.  This rule was definitely a big pain to try implement correctly when I first worked on it.

internal/rules/prefer_optional_chain/prefer_optional_chain_test.go
```diff
@@ -983,6 +983,12 @@ func TestPreferOptionalChainRule(t *testing.T) {
 				return obj && obj?.flag === false;
 			}
 		`},
+		{Code: `declare const foo: { bar: string | null }; foo.bar === null || foo.bar.trim() === '';`},
+		{Code: `'bar' in foo[0] && foo[0].bar === 'err';`},
```
connorshea: This one fails in [the playground](https://typescript-eslint.io/play#ts=5.9.3&fileType=.tsx&code=PTAEBMFMGMBsEMBOlTQPYDsDOAXUAzNNALlAG9QAjJU3RASwwHNQAfUDAV1llAF8A3ASIA6aolABeaR269W7QmjFIROBgFsAFAEopMgOQGBAKBCgD4g6EbC0AbQAMAXVAAyN3afOVE6ZItIRERjMzAoOCQUdGw8JVIAZUgcAB46RiYAPjZQTgwofEZIcCEcAE8AB0g0fDt9AIM8gqLwawU7ESx6AC8Uf1BHU3MIhGRUTFw7Unt05gAaXIwAaww0AHcMe2dnLZymyEKMYtLK6tqleot9w%2BK2xVFYSGYcAAtLwbCIGFHoibiiUiUIiPeAYIQjKLjWJUGjkUAAN3gsE4kFo6gyOS4PH4Qgu7XEIkRyL6hgAkhgifRwKAACLwHCQAAq9A0kDuMMQhKRKMuWNgpghYxik3icPEaIYzH4e3yBxaJyqNTq-UaspurTY92U4gEQA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Y6RAM0WloHsalfkwCG8WmQAWo5uii9o-aJHBgAviHVA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false) due to being unable to resolve foo, but otherwise it looks like all of these do pass on the original rule.

kalvenschraut: updated the test to declare foo

kalvenschraut:
image
Did double check that the tests fail on main for all the ones I added

PR #739, opened 2026-02-24
feat: add env var to suppress program diagnostics

### LINKED ISSUE #736: Support disabling/suppressing `tsconfig-error` rule
I have been working on migrating Sentry's codebase to use oxlint but the migration path has a hard dependency on type awareness and typescript version.

**The issue**

Since we are staying on 5.x for the time being we would like to still use oxlint but disable the tsconfig verification, which is mostly fired for old deprecated options like `alwaysStrict` and other changes like `baseDir` and `rootDir`.

Turning the rule off via cli or oxlint config doesn't work because it seems like it's not a "real" rule.  It's produced during program creation when TypeScript fails to parse or validate a `tsconfig.json` unlike the other rules.

**Suggestion/Feature Request**

I'm thinking large codebases can benefit from this by silencing this type of errors temporarily until they are able to migrate to TS 6.0/7.0 since the `tsconfig` doesn't really stop those rules from working.

I cobbled something together locally by adding an environment variable: `TSGOLINT_REPORT_TSCONFIG_ERRORS=false`, this works today without any changes on the oxlint side. It's technically "suppressing" it rather than "disabling it".

I considered adding top-level payload flag but that requires changes on `oxlint` side to pass it along, so env variable was the easiest option to test with and is "secret" enough to not encourage everyone adopting it.

What do you folks think? I'm happy to PR this if you think it's worth adding.

logaretm:
Adds `OXLINT_TSGOLINT_SUPPRESS_PROGRAM_DIAGNOSTICS` env variable check to suppress errors coming out of `tsconfig-error` diagnostics.

Usage:

```bash
OXLINT_TSGOLINT_SUPPRESS_PROGRAM_DIAGNOSTICS=true pnpm oxlint . --type-aware
```

closes #736


cmd/tsgolint/headless.go
```diff
@@ -398,6 +398,9 @@ func runHeadless(args []string) int {
 			diagnosticsChan <- ruleToAny(d)
 		},
 		func(d diagnostic.Internal) {
+			if d.Id == "tsconfig-error" && os.Getenv("OXLINT_TSGOLINT_SUPPRESS_PROGRAM_DIAGNOSTICS") == "true" {
```
camc314: `os.Getenv` isn't a cheap call - can we move this out of this function callback into the `runHeadless` scope so it's only ever called once

logaretm: Thanks, done!

logaretm:
> I don't think this will work how you expect...

@camc314 I didn't realize that, the issue I guess is with invalid config the program produced is `nil` so I guess we are forced to skip them?

logaretm:
Never mind, I think I misunderstood what happens in that case. I think we can just ignore `configParseResult.Errors` when the flag is enabled.

I pushed a re-work of this that has the desired effect but now it also lints those projects files.

logaretm:
@camc314 Any chance we can take another look at this?

e2e/snapshot.test.ts
```diff
@@ -532,6 +532,24 @@ console.log(x);
     expect(diagnostics).toMatchSnapshot();
   });
 
+  it('should suppress tsconfig-error diagnostics when OXLINT_TSGOLINT_SUPPRESS_PROGRAM_DIAGNOSTICS env var is true', async () => {
```
camc314: there should be a test making sure that we report some sort of type aware lint rules when the flag is enabled.

internal/linter/linter.go
```diff
@@ -63,15 +64,17 @@ func RunLinter(
...
 
-		if len(diagnostics) > 0 {
-			for _, d := range diagnostics {
-				onInternalDiagnostic(d)
+		if program == nil {
+			if len(diagnostics) > 0 {
+				for _, d := range diagnostics {
+					onInternalDiagnostic(d)
+				}
 			}
```
camc314: this length check feels unnecessary?

logaretm: Nope, I will remove it.

logaretm:
Thanks for the review, I addressed both comments 🙏

logaretm:
Weird, I didn't get those errors when I ran `just ready` initially, after rebasing I got the same errors as the CI.

I fixed them now.

internal/utils/create_program.go
```diff
@@ -36,6 +36,10 @@ func CreateProgram(singleThreaded bool, fs vfs.FS, cwd string, tsconfigPath stri
 	configParseResult, diagnostics := tsoptions.GetParsedCommandLineOfConfigFile(tsconfigPath, &core.CompilerOptions{}, nil, host, nil)
 
 	if len(diagnostics) > 0 {
+		// configParseResult is nil here — the file couldn't be read at all
+		if suppressProgramDiagnostics {
+			return nil, nil, nil
+		}
```
camc314: I'm going to revert this change - it should never happen

PR #626, opened 2026-02-05
feat: add `--stats` flag for profiling

no-yan:
Resolves #614 

Some changes from the initial issue:
1. Removed CPU time of `typecheck`, because we cannot get it without patching tsgo internals.
2. removed wall time of lint, because rules run interleaved across files, and showing elapsed time is meaningless.
3. Add "Traversal+overhead" in lint phase to detect non-rule overhead.

Implement note:
- Measure worker CPU time to calc non-rule overhead.
- Move nil checks inside metric functions to improve readability. Doing nil checks at the caller side made the code verbose. However, ruleListener keeps the check at the call site to minimize function calls, as it is a hot path.

### Output

```
Typecheck:
    Program                                                  Wall Time     Files
    tsconfig.json                                               0.061s        16
    inferred program                                            0.012s         3
    ────────────────────────────────────────────────────────────────────────────
    Total                                                       0.072s

Lint:
    Rule                                                      CPU Time
    await-thenable                                              0.087s
    unbound-method                                              0.062s
    no-floating-promises                                        0.041s
    no-base-to-string                                           0.011s
    no-duplicate-type-constituents                              0.001s
    ... 6 more rules (0.000s)
    ──────────────────────────────────────────────────────────────────
    Traversal+overhead                                          0.002s
    Total                                                       0.205s

Summary:
    Category                                                 Wall time
    typecheck                                                   0.072s
    lint                                                        0.071s
```

PR #740, opened 2026-02-24
feat(headless): emit per-rule timing data

wagenet:
## Summary

- When `TIMING` env var is set, wraps each rule listener with a nanosecond timer
- After all diagnostics are written, sends a `headlessMessageTypeTiming` (type 2) binary-framed message with payload `{"timings": {"rule-name": ms_float, ...}}`
- The Rust consumer (oxc) reads this message and merges per-rule durations into `TimingStore`, allowing them to appear in the `--timing` output alongside built-in rule timings

## Test plan

- [ ] Build tsgolint and run `TIMING=1 oxlint --timing --type-aware <files>` — type-aware rules should appear in the timing table
- [ ] Without `TIMING` set, no timing message is sent (existing behavior unchanged)

### DIFF
--- a/cmd/tsgolint/headless.go
+++ b/cmd/tsgolint/headless.go
@@ -11,6 +11,7 @@ import (
 	"runtime"
 	"slices"
 	"sync"
+	"time"
 
 	"github.com/go-json-experiment/json"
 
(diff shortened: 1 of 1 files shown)


camc314:
closing for now - see my comment on the oxc PR


Also, I think we'd just want something more like https://github.com/oxc-project/tsgolint/pull/626

PR #751, opened 2026-02-26
fix(return-await): guard concise arrow body path with hasAsync check

wagenet:
## Summary

- Non-async concise arrow functions (e.g. `const f = (): Promise<number> => Promise.resolve(1)`) were incorrectly flagged in `always`, `in-try-catch`, and `error-handling-correctness-only` modes
- The `KindArrowFunction` exit handler called `testEachPossiblyReturnedNode` on the concise body without checking `scope.hasAsync`, bypassing the guard that already exists on the `KindReturnStatement` handler
- Fix adds `&& scope != nil && scope.hasAsync` to the concise-body branch, making it consistent with the block-body path

## Test plan

- [ ] Added three valid (no-error) test cases for a non-async typed arrow returning a `Promise`, one per affected mode (`always`, `in-try-catch`, `error-handling-correctness-only`)
- [ ] `go test ./internal/rules/return_await/...` passes


internal/rules/return_await/return_await_test.go
```diff
@@ -228,6 +228,18 @@ async function test(unknownParam: unknown) {
...
+			Code:    "const f = (): Promise<number> => Promise.resolve(1);",
+			Options: rule_tester.OptionsFromJSON[ReturnAwaitOptions](`"always"`),
+		},
+		{
+			Code:    "const f = (): Promise<number> => Promise.resolve(1);",
+			Options: rule_tester.OptionsFromJSON[ReturnAwaitOptions](`"in-try-catch"`),
+		},
+		{
+			Code:    "const f = (): Promise<number> => Promise.resolve(1);",
+			Options: rule_tester.OptionsFromJSON[ReturnAwaitOptions](`"error-handling-correctness-only"`),
```
connorshea, line 241: Confirmed that all three of these pass in [typescript-eslint playground](https://typescript-eslint.io/play#ts=5.9.2&fileType=.tsx&code=MYewdgzgLgBAZjAvDAFASgFwwAoCcQC2AlhAKYA8YArgQEam4B8SzehJpAdLqRCADYA3UigCMaANxA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6aRfWaJ2gQwDufSvnRgA2pETRoAe2iQANFGa180QrTJ98ZABaQAuuDABfEKaA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false)

camc314: Thanks connor!

PR #753, opened 2026-02-26
fix(prefer-optional-chain): avoid false positives for instanceof and redundant nullish checks

wagenet:
## Summary

- **`instanceof` false positive**: patterns like `(!a.b || foo instanceof a.b)` and `(a.b && foo instanceof a.b)` were incorrectly flagged. Added `KindInstanceOfKeyword` to the invalid-operator switches in both the AND-chain and OR-chain branches of `parseOperand` — `instanceof` has no optional-chain equivalent.

- **Redundant nullish-check false positive**: `a.b === null || a.b === undefined` was falsely reported because both operands check the exact same expression with no deeper member access to convert. Added an `allOperandsCheckSameExpression` guard to `validateOrChainForReporting`, mirroring the identical guard already present in `validateAndChainForReporting`.

## Test plan

- [ ] `(!a.b || foo instanceof a.b)` — no violation
- [ ] `(a.b && foo instanceof a.b)` — no violation
- [ ] `request.payload === null || request.payload === undefined` — no violation
- [ ] `request.payload === undefined || request.payload === null` — no violation
- [ ] `go test ./internal/rules/prefer_optional_chain/...` passes

### DIFF
--- a/internal/rules/prefer_optional_chain/prefer_optional_chain_test.go
+++ b/internal/rules/prefer_optional_chain/prefer_optional_chain_test.go
@@ -4173,9 +4173,6 @@ foo.bar?.() === undefined || foo.bar?.().baz;
 		MutateOutput: AddTrailingEqualUndefined(Identity),
 	})...)
 
-	// --- `in` / `instanceof` operator false positives ---
-	// `in` has no optional-chain equivalent (`key in obj?.prop` is not valid),
-	// so these patterns must never be reported.
 	validCases = append(validCases, rule_tester.ValidTestCase{
 		Code: `(!data.previous_values || key in data.previous_values)`,
 	}, rule_tester.ValidTestCase{
@@ -4189,10 +4186,6 @@ foo.bar?.() === undefined || foo.bar?.().baz;
 	}, rule_tester.ValidTestCase{
 		Code: `(a.b && foo instanceof a.b)`,
 	})
-
-	// --- Redundant null/undefined checks false positive ---
-	// `a.b === null || a.b === undefined` is just a nullish check — there is no
-	// member access to convert, so it must not be reported.
 	validCases = append(validCases, rule_tester.ValidTestCase{
 		Code: `request.payload === undefined || request.payload === null`,
 	}, rule_tester.ValidTestCase{


camc314:
@wagenet mind rebasing? thanks

kalvenschraut:
May want to double check main has these fixed, looks pretty similar to my other PR that was merged.  AI wanted to add instanceof also but those weren't actually being triggered at least the ones I tried 

https://github.com/oxc-project/tsgolint/pull/732

internal/rules/prefer_optional_chain/prefer_optional_chain.go
```diff
@@ -2550,6 +2552,12 @@ func (processor *chainProcessor) validateOrChainForReporting(chain []Operand) []
 		return nil
 	}
 
+	// When every operand checks the exact same expression (e.g., a.b === null || a.b === undefined),
+	// there is no member access to convert — skip flagging, mirroring the AND-chain guard.
+	if processor.allOperandsCheckSameExpression(chain) {
```
kalvenschraut: this is added already  just a few lines down https://github.com/oxc-project/tsgolint/pull/753/changes#diff-0d230715551b2aebfce0eb9756db9dab5b27e4784402ce304b55bcf884976522L2560

internal/rules/prefer_optional_chain/prefer_optional_chain_test.go
```diff
@@ -4166,6 +4166,32 @@ foo.bar?.() === undefined || foo.bar?.().baz;
 		MutateOutput: AddTrailingEqualUndefined(Identity),
 	})...)
 
+	// --- `in` / `instanceof` operator false positives ---
+	// `in` has no optional-chain equivalent (`key in obj?.prop` is not valid),
+	// so these patterns must never be reported.
+	validCases = append(validCases, rule_tester.ValidTestCase{
+		Code: `(!data.previous_values || key in data.previous_values)`,
+	}, rule_tester.ValidTestCase{
```
kalvenschraut, line 4178: I cloned this file from this branch and all these test cases pass on main so I think my changes in #732 are sufficient

wagenet: Huh... that doesn't match what I was seeing. Guess we'll need to dig in more.

kalvenschraut: Was your main up to date? As my PR was recently merged/released

wagenet: Yes, I updated it after I rebased.

kalvenschraut: https://github.com/user-attachments/assets/f6273823-0d84-4170-b23f-83e6c69c74e0

above shows what I am doing at least 🤷  I started with your changes locally as shown by the diff and then show that i have no other changes and I am on main. Then I run the test and they still pass meaning no changes are required to satisfy these test cases. Unless I am overlooking something

camc314 (approved):
Thanks!

PR #768, opened 2026-03-06
Revise oxlint command examples in README

connorshea:
I don't think we should encourage using `--type-check` by default as it can be confusing for users on TypeScript versions <6 when they get different results, or errors. Which is going to be the vast majority of users. Plus AI will just assume it should always do this, which would be bad in these cases.


Boshen:
Good call.

PR #784, opened 2026-03-10
fix(no-unnecessary-condition): handle IndexedAccess types in ?? and ??= paths

younggglcy:
## Summary

- Fix false positive in `no-unnecessary-condition` when the left side of `??` or `??=` is an unresolved indexed access type (e.g., `Partial<Record<R, T[]>>[R]` where `R` is a generic type parameter)
- Resolve the type via `getBaseConstraintOfType`; if the constraint is still unresolvable or includes nullish, skip the report conservatively
- Extract shared logic into a `checkIndexedAccessNullish` closure to avoid duplication between `??` and `??=` paths

Related Issue in oxc: https://github.com/oxc-project/oxc/issues/20149

## Test plan

- Added two valid test cases for `Partial<Record<R, T[]>>` (with and without `noUncheckedIndexedAccess`)
- Verified existing `invalid-49` test (`T[K]` with known non-nullish constraint) still correctly reports
- All existing tests pass: `go test ./internal/rules/no_unnecessary_condition/ -count=1`

### DIFF
--- a/internal/rule_tester/__snapshots__/no-unnecessary-condition.snap
+++ b/internal/rule_tester/__snapshots__/no-unnecessary-condition.snap
@@ -1893,3 +1893,21 @@ Message: Unnecessary conditional, value is always truthy.
      |                ~~~~~~~
    5 | }
 ---
+
+[TestNoUnnecessaryConditionRule/invalid-136 - 1]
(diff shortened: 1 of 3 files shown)


camc314 (approved):
thank you!

Issue #810, opened 2026-03-18
SIGSEGV in NameResolver.Resolve when `.await` is used as a property name


turadg:
## Description

Type-aware linting crashes with a nil pointer dereference in `NameResolver.Resolve` when a file accesses `.await` as a property on an imported binding, in a file that also uses a JSDoc `@import` type annotation.

The crash is triggered by the `await` keyword being used as a property name via dot notation. The name resolver dereferences a nil scope/container for the `await` identifier node. Other keyword-as-property names (e.g., `.return`) do not trigger the crash.

## Motivation

This pattern is common in the [`@endo/patterns`](https://github.com/endojs/endo/tree/master/packages/patterns) library, where `M.await()` is a standard API for describing async method guard shapes:

```js
// from @endo/patterns patternMatchers.js
const M = freeze({
  // ...
  await: argPattern => makeAwaitArgGuard(argPattern),
  // ...
});
```

Consumer code like this is typical across the Agoric SDK (~29 call sites):

```js
M.interface('Foo', {
  getPublicFacet: M.callWhen(M.await(InstanceHandleShape)).returns(M.remotable()),
});
```

## Minimal reproduction (no third-party dependencies)

```js
import { join } from 'node:path';
/** @import {X} from 'y'; */
const x = join.await;
/** @type {X} */
let z;
```

Both conditions are required:
1. `.await` property access on an **imported** binding (local objects do not trigger the crash)
2. A JSDoc `@import` type that is **used** as a type annotation somewhere in the file

Removing either condition prevents the crash.

## Stack trace

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
[signal SIGSEGV: segmentation violation code=0x2 addr=0x40]

github.com/microsoft/typescript-go/internal/binder.(*NameResolver).Resolve(...)
    typescript-go/internal/binder/nameresolver.go:96
github.com/microsoft/typescript-go/internal/checker.(*Checker).resolveEntityName(...)
    typescript-go/internal/checker/checker.go:15225
github.com/microsoft/typescript-go/internal/checker.(*Checker).getSymbolOfNameOrPropertyAccessExpression(...)
    typescript-go/internal/checker/checker.go:30830
github.com/microsoft/typescript-go/internal/checker.(*Checker).getSymbolAtLocation(...)
    typescript-go/internal/checker/checker.go:30595
github.com/microsoft/typescript-go/internal/checker.(*Checker).GetSymbolAtLocation(...)
[...]

PR #811, opened 2026-03-18
fix: recover from upstream panics in rule listeners

turadg:
Prompted by #810.

Adds defer/recover in runListeners so that upstream panics (e.g. typescript-go nil dereference on keyword property names like .await) are caught gracefully instead of crashing the linter.

AI-Assist: Opus 4.6


camc314:
Closing as not planned - I may come back to this, but I'm worried about the perf overhead of this.

Furthermore, the stack trace alone is usually enough information to fix the panic.

PR #823, opened 2026-03-23
fix: prevent tsconfig exclude files from checking

liangmiQwQ:
If we have a `tsconfig.json` with `exclude` like this

```json
{
  "include": ["src", "index.ts"]
  "exclude": ["a.ts", "main.ts"]
}
```

But if we run `oxlint --type-aware --type-check`, the files in `exclude` will still be checked because of the inferred program added in #140. That is unexpected, we should follow the behavior where tsc won't check files which don't match any `tsconfig`

~~This PR removed the creation logic of inferred program, it's actually a revert of #140 and the test modify of #215~~

Partial related to voidzero-dev/vite-plus#1104

### DIFF
--- a/e2e/__snapshots__/snapshot.test.ts.snap
+++ b/e2e/__snapshots__/snapshot.test.ts.snap
@@ -15,37 +15,6 @@ See https://github.com/oxc-project/tsgolint/issues/351 for more information.",
 ]
 `;
 
-exports[`TSGoLint E2E Snapshot Tests > should correctly lint files not inside a project 1`] = `
-[
-  {
-    "file_path": "fixtures/with-unmatched-files/src/index.ts",
-    "kind": 0,
-    "message": {
-      "description": "Unsafe argument of type any assigned to a parameter of type string.",
-      "id": "unsafeArgument",
-    },
-    "range": {
-      "end": 45,
-      "pos": 36,
-    },
-    "rule": "no-unsafe-argument",
-  },
-  {
-    "file_path": "fixtures/with-unmatched-files/test.ts",
-    "kind": 0,
-    "message": {
-      "description": "Unsafe argument of type any assigned to a parameter of type string.",
-      "id": "unsafeArgument",
-    },
-    "range": {
-      "end": 45,
-      "pos": 36,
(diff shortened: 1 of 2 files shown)


camc314 (changes requested):
>  That is unexpected, we should follow the behavior where tsc won't check files which don't match any tsconfig

This is incorrect - it's expected behaviour that we typecheck all files, regardless of whether the match any tsconfig

We do this to provide a seamless, no-config user experience rather than having users fiddle with tsconfigs and try to work out how it all fits together

The real bug here is that we create 1x inferred program for all files that don't have a tsconfig, this causes the node types to "leak" hence sometimes you see the FP, and sometimes you don't depending on whether `vite.config.ts` is linted (`vite.config.ts` has a `///references node types` in one of the declaration files it imports

liangmiQwQ:
Well... I got it. I can totally understand how important it is to provide an easy use type checker for simple or single file project's linting.

However, besides the type leak issue, for large projects, or which just migrated to `oxlint`, if the users still want to use `oxlint` or `vp check` with typecheck instead of a separated `tsc`, I think it's still meaningful to respect tsconfig's `exclude` / `include`. (Follow [the behavior in typescript-eslint](https://github.com/typescript-eslint/typescript-eslint/blob/e9cc25a794dbf82a241807f394489a2aad39ffae/packages/typescript-estree/src/create-program/createProjectProgramError.ts#L104) ([docs](https://tseslint.com/none-of-those-tsconfigs-include-this-file)))

These two requests are not actually in conflict. Perhaps an acceptable solution is to only create the inferred program for no tsconfig project (all files are unmatched) or provide a configurable option, means typecheck will still obey the options like `exclude / include` in the existing tsconfig while still keep the typecheck usable for no-config users

liangmiQwQ:
I've pushed [several commits in this branch](https://github.com/liangmiQwQ/tsgolint/commit/bb08656f5230289bfee7b162a33a4cb566446c4c) about the solution I mentioned above, if you also think it is acceptable and reasonable for tsgolint, you can consider to reopen it. /cc @camc314

PR #836, opened 2026-03-26
feat: add no-object-comparison rule

### LINKED ISSUE #746: New rule proposal: `no-object-comparison`
We have this custom eslint rule to prevent common mistakes while working with BigNumbers (or any other custom classes representing numbers, dates or really anything else):

Some invalid code examples:

```
import { BigNumber } from 'bignumber.js'
const a = BigNumber(10)
const b = BigNumber(20)

// this is mistake because most likely you want to use `.eq`
if (a === b) {
  a.plus(1)
}

// this is always mistake because one should use `.gte`
if (a >= b) {
  a.plus(1)
}
```

This rule (in our design) requires providing explicit list of class names that will trigger validation: `options: [{ classNames: ['BigNumber'] }],`. I can imagine that using many different binary operators (`>`, `<`) is always (?) incorrect for custom types while `===` might be sometimes valid (hence `classNames` configuration). So we might want to tweak this...

Please let me know if you would like for us to work on PR with such rule. It's pretty essential while working a lot with custom numeric types and it's literally one of the last two rules that we run eslint for ;)

yivlad:
Closes #746

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,13 @@ Implemented 59/61.
 - [x] [unbound-method](https://typescript-eslint.io/rules/unbound-method)
 - [x] [use-unknown-in-catch-callback-variable](https://typescript-eslint.io/rules/use-unknown-in-catch-callback-variable)
 
+## Custom Rules
+
+- [x] `no-object-comparison`
+      Prevents direct comparisons on configured object types. By default it forbids
+      relational operators (`<`, `<=`, `>`, `>=`), and
+      `forbidEqualityOperators: true` also forbids `==`, `===`, `!=`, and `!==`.
+
 ## Links
 
 - [ARCHITECTURE.md](./ARCHITECTURE.md) - Detailed technical documentation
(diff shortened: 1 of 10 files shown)


yivlad:
Hey @camc314, is there any chance we could merge this?

camc314:
Hi, unfortunately i'm not planning to merge in the immediate future - any additional rules outside of the typescript-eslint scope at this time - we're currently working hard on oxlint, to support vite+, and planning to support better support for linting framework files (vue svelte etc).

yivlad:
@camc314 do you maybe know if we can achieve similar functionality as in this PR by using some plugins?

krzkaczor:
@camc314 I ask you to reconsider. This rule is really critical when working with BigNumber like libraries which are very common when dealing with financials. Same goes for the whole Effect ecosystem (custom date, decimals types). 

Here's a thread on eslint rule: https://github.com/typescript-eslint/typescript-eslint/issues/5647

camchenry:
Thank you for your contribution. I agree with @camc314: we do not currently have the capacity to support adding new rules beyond what typescript-eslint already supports. I've added a note about this in https://github.com/oxc-project/tsgolint/pull/972.

We hope in the future to add support for type checking in oxlint in the future that will enable rules like this to exist as plugins, but that is a long way off still. https://github.com/oxc-project/oxc/issues/19596

PR #838, opened 2026-03-27
fix: false positive on conditional types from mutually recursive type aliases

cmd8:
## Problem

`no-unsafe-assignment` reports "Unsafe assignment of an error typed value" on types that use mutually recursive conditional type aliases imported from `.d.ts` files. `tsc 5.9.3` has no errors on the same code.

The pattern that fails:

```typescript
// in node_modules/pkg/index.d.ts
type Default<T extends {}, U> = [keyof T] extends [never] ? U : T;
type SubResponse<T extends Opts> = { items: ResultComponent<T>[] };    // ← references ResultComponent
type ResultComponent<T extends Opts> = Default<
    (T["a"] extends object ? { a: string } : {}) &
    (T["b"] extends number ? SubResponse<T> : {}),                     // ← references SubResponse
    { a: string }
>;
export type Result<T extends Opts> = { id: string } & ResultComponent<T>;

// in user code
declare const r: Result<{ a: { x: number } }>;
const id: string = r.id;  // error: "Unsafe assignment of an error typed value"
```

`tsc --noEmit` passes. `tsgo --noEmit` reports `TS2456: Type alias 'ResultComponent' circularly references itself`.

Found in the wild with `exa-js@1.8.23` whose `SearchResult<T>` type uses this pattern through `ContentsResultComponent<T>` and `SubpagesResponse<T>`.

## Cause

During type alias resolution, the checker instantiates `Default<intersection, U>`. The check type `[keyof T]` is instantiated with `T` mapped to the intersection of deferred conditional types. This calls `getIndexType(intersection)`, which calls `getReducedType(intersection)`.

`getReducedType` calls `getPropertiesOfUnionOrIntersectionType` to check if the intersection has conflicting discriminant properties (never-reduction). This resolves the apparent type of each deferred conditional type, which evaluates both branches of each conditional, which reaches `SubResponse<T>`, which needs `ResultComponent<T>`, which is already on the type resolution stack. The checker treats this as a circular reference and returns the error type.

## Fix

In `getReducedType`, skip the never-intersection property check when the intersection contains instantiable types (deferred conditional types, type parameters). Properties of unresolved types cannot produce conflicting discriminants, so the check is not needed. The reduction happens correctly later when concrete type arguments are provided.

## Test

Added a test case to `no-unsafe-assignment` that simulates the `exa-js` type pattern with virtual `node_modules` files containing mutually recursive conditional type aliases.

### DIFF
--- a/patches/0007-fix-skip-never-intersection-reduction-for-instantiab.patch
+++ b/patches/0007-fix-skip-never-intersection-reduction-for-instantiab.patch
@@ -1,0 +1,41 @@
+From 9c40f6d5aeb049c9ab85ed2ca3d7cf1cbf610f0d Mon Sep 17 00:00:00 2001
+From: Eugene Oldman <eugene.oldman@gmail.com>
+Date: Fri, 27 Mar 2026 11:54:42 +1100
+Subject: [PATCH] fix: skip never-intersection reduction for instantiable types
+
+When getReducedType is called on an intersection containing instantiable
+types (deferred conditional types, type parameters), skip the property
+check that determines if the intersection reduces to never. Resolving
+properties of instantiable types can trigger circular type resolution
+when mutually recursive type aliases reference each other through
+conditional types.
+---
+ internal/checker/checker.go | 12 ++++++++++--
+ 1 file changed, 10 insertions(+), 2 deletions(-)
+
+diff --git a/internal/checker/checker.go b/internal/checker/checker.go
+index 5bf47a104..49622be6c 100644
+--- a/internal/checker/checker.go
++++ b/internal/checker/checker.go
+@@ -21225,8 +21225,16 @@ func (c *Checker) getReducedType(t *Type) *Type {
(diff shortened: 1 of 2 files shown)


camc314 (changes requested):
> The pattern that fails:

In your PR description, `Opts` is undefined - can you define it to help me understand the issue?

Furthermore, I tried looking at the package source, and it looked different to the code in the description https://npmx.dev/package-code/exa-js/v/1.8.23/dist%2Findex.d.ts

patches/0007-fix-skip-never-intersection-reduction-for-instantiab.patch
camc314, line 1: I'm incredibly concered by this change, and i'm unconvinced this is the correct fix.

Can you make a minimal reproduction of the issue for me to have a closer look at?

Thanks

camc314:
Closing as I don't have enough detailed for a concrete reproduction hence we can't fix this issue.

Please make an issue to discuss the problem - thanks

PR #914, opened 2026-04-27
fix(prefer-optional-chain): avoid access comparison false positive

### LINKED ISSUE #21849: linter: avoid `prefer-optional-chain` breaking symmetry

1.61.0

### What command did you run?

oxc neovim lsp

complicated generated config at @framer

### What happened?

More of a rule improvement than a bug to prevent certain false positives.

```
function test() {
	let x: undefined | { a: string }, y: undefined | { a: string }
	if (x && y && x.a === y.a) {
		console.log("x and y are equal")
	}
}
```

`prefer-optional-chain` flags this `if`, and it does correctly stop flagging when converting one of `x` or `y` to optional chaining (to prevent `undefined === undefined`).
But it's not ideal to flag at all because we don't want to break the symmetry here for human readability.
One way that this case can be prevented/detected (even without extra type info checks) if I'm not mistaken is e.g. by recursively applying the rule:

Apply rule to whole condition, do not stop after detecting one flag. IF rule flags `x` *and* rule flags `y` AND in the modified tree after autofixing `x` it doesn't flag `y` anymore AND in the modified tree after autofixing `y` it doesn't flag `x` anymore. (mutual exclusivity)
In that case, we don't want either flagged in the first place.

Idk if y'all have the infra to implement that easily, but wanted to suggest the improvement in any case. There might also be an easier way to detect it than my spontaneously conceived algo.

camc314:
Prevents `prefer-optional-chain` from reporting access-vs-access comparisons like `x && y && x.a === y.a`, where optional chaining would change when the other side is evaluated.

Adds a regression test for the valid guarded comparison case; fixes https://github.com/oxc-project/oxc/issues/21849.

### DIFF
--- a/internal/rules/prefer_optional_chain/prefer_optional_chain.go
+++ b/internal/rules/prefer_optional_chain/prefer_optional_chain.go
@@ -1533,14 +1533,11 @@ outer3:
 		leftIsAccess := utils.IsAccessExpression(left)
 		rightIsAccess := utils.IsAccessExpression(right)
 
-		// If both sides are access expressions with the same base, we can't convert
-		// to optional chaining (e.g., foo && foo.bar === foo.baz)
+		// If both sides are access expressions, we can't convert to optional
+		// chaining without changing when the other side is evaluated.
+		// For example, `x && y && x.a === y.a` must keep both guards.
 		if leftIsAccess && rightIsAccess {
-			leftBase := getBaseIdentifier(left)
-			rightBase := getBaseIdentifier(right)
-			if areNodesStructurallyEqual(leftBase, rightBase) {
-				return Operand{typ: OperandTypeInvalid, node: node}
-			}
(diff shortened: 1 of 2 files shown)


camc314:
### Merge activity

* **Apr 27, 2:45 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Apr 27, 2:45 PM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Apr 27, 2:52 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

PR #942, opened 2026-05-10
fix(no-unnecessary-type-assertion): add mutual assignability fallback

johnsoncodehk:
## Summary

`isTypeUnchanged` currently relies solely on pointer equality (`uncast == cast`) to determine if a type assertion is unnecessary. This misses cases where the TypeScript checker creates structurally identical but pointer-distinct type objects — for example, a type resolved through property access vs the same type alias referenced directly in an `as` annotation.

This PR adds a bidirectional `isTypeAssignableTo` check as a fallback (before the `ExactOptionalPropertyTypes` path), matching the behavior of typescript-eslint's [`areMutuallyAssignable`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts) helper.

## Reproduction

```typescript
interface Item { data: VueCompletionData }
declare const item: Item;
const d = item.data as VueCompletionData; // should report: assertion is unnecessary
```

The checker returns distinct `*checker.Type` pointers for `item.data`'s resolved type and the `VueCompletionData` reference in the `as` clause, even though they are the same logical type. The pointer check fails → false negative.

## Fix

```go
if checker.Checker_isTypeAssignableTo(ctx.TypeChecker, uncast, cast) &&
    checker.Checker_isTypeAssignableTo(ctx.TypeChecker, cast, uncast) {
    return true
}
```

This is the same API already used by `no-unsafe-type-assertion`, `prefer-reduce-type-parameter`, and `related-getter-setter-pairs` in this repo.

### DIFF
--- a/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
+++ b/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
@@ -1813,5 +1813,35 @@ const s2 = (s);
 				},
 			},
 		},
+		// Mapped types like Pick<T, keyof T> produce structurally identical but
+		// pointer-distinct types. The checker returns different type objects for
+		// `Data` and `Same<Data>` even though they are mutually assignable.
+		// Currently fails: isTypeUnchanged uses pointer equality only.
+		// Fixing this requires isTypeAssignableTo, but that is too lenient in
+		// typescript-go (e.g. array↔tuple, literal↔enum member), causing
+		// false positives on valid assertions. Blocked on typescript-go.
+		//
+		// {
+		// 	Code: `
+		// interface Data { name: string; value: number }
+		// type Same<T> = Pick<T, keyof T>;
+		// declare const item: Data;
+		// const d = item as Same<Data>;
+		//       `,
+		// 	Output: []string{`
+		// interface Data { name: string; value: number }
+		// type Same<T> = Pick<T, keyof T>;
+		// declare const item: Data;
+		// const d = item;
+		//       `,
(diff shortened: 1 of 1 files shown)


camc314 (changes requested):
Hi, thanks for putting up this PR.

I tried this locally, and commented out your changes in `internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go` and the tests still passed.

Could you make a minimal repro of the issue first, and make sure that the test is failing before fixing?

Thanks!

johnsoncodehk:
After investigating, I found that:

1. The original test case (simple property access) doesn't actually reproduce the issue — the Go checker already returns pointer-equal types there, as @camc314 pointed out.

2. The real reproduction is **mapped types** like `Pick<T, keyof T>`: the checker returns pointer-distinct but mutually assignable types for `Data` vs `Same<Data>`.

3. However, adding a bidirectional `isTypeAssignableTo` fallback causes **7 valid test cases to regress** because `typescript-go`'s `isTypeAssignableTo` is too lenient compared to TypeScript's — it incorrectly considers these pairs mutually assignable:
   - `(string | number)[]` ↔ `[number, string]` (array vs tuple)
   - `[3, 'hi', 'bye']` ↔ `Tuple` (literal tuple vs type alias)
   - `{}` ↔ `PossibleTuple` (object literal vs type alias)
   - `1` ↔ `T.Value2` (numeric literal vs enum member)

This is blocked on `typescript-go` fixing its assignability checks. I've pushed a commented-out test case documenting the repro. Closing for now.

PR #967, opened 2026-05-18
perf(no-unnecessary-type-parameters): stop counting settled candidates

### LINKED ISSUE #961: `typescript/no-unnecessary-type-parameters` rule is quite slow
I had Claude create a very simple benchmark of the vscode repo to confirm this, and it definitely seems that the rule is by far the slowest (or at least one of the slowest) in tsgolint. (note that the `--debug timings` was not used for this)

See [here](https://github.com/connorshea/vscode/blob/b3b09ce9accadf2b336ee58012a2f91d273196d4/BENCHMARK-RESULTS.md) for benchmark configs and more notes.

The following is results from Claude running the benchmarks (3 runs each with hyperfine):

----

## Results

`oxlint -c <config> src/`, all configs with `options.typeAware: true`. The
"Relative" column is normalized to the fastest config.

| Config | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| no type-aware rules                                  | 16.31 ± 1.09 | 15.66 | 17.57 | 1.00x |
| `await-thenable` only                                | 17.01 ± 1.11 | 15.74 | 17.76 | 1.04x |
| `no-unnecessary-type-parameters` only                | 45.20 ± 0.96 | 44.10 | 45.84 | 2.77x |
| `await-thenable` + `no-unnecessary-type-parameters`  | 44.21 ± 0.68 | 43.69 | 44.99 | 2.71x |

- Enabling `no-unnecessary-type-parameters` adds **~28.9 s** over a run with no type-aware rules.
- `await-thenable` alone adds <1 s
- Adding `no-unnecessary-type-parameters` on top of `await-thenable` costs about the same **+27 s**, so the cost is the rule itself, not tsgolint startup.

camc314:
Improves `no-unnecessary-type-parameters` performance by avoiding the expensive recursive type-usage walk for type parameters that are already obviously safe to skip.

The rule already had an AST-based fast path:

 ```go
isTypeParameterRepeatedInAST(...)
```

but it ran after countTypeParameterUsage, which is the expensive checker-heavy part. This meant we were paying the full recursive type analysis cost even for type parameters that could be ruled out cheaply.

This PR moves that cheap filtering earlier, then narrows the expensive pass to only the remaining candidates.

## Why This Is Faster

The slow path is countTypeParameterUsage, which recursively walks TypeScript types and calls into the checker. On large projects like VS Code, most generic declarations are not reportable because their type parameters are used more than once in the signature.

Before this change, the rule did this for every generic declaration:

 1. recursively count type parameter usage with checker types
 2. collect AST references
 3. skip if the AST already proves the type parameter is repeated

That order wastes work. If the AST can prove the type parameter appears multiple times before the function body, then the rule will not report it, so there is no reason to do the recursive checker walk first.

After this change, the rule does this instead:

 1. collect AST references
 2. keep only type parameters that are still possible reports
 3. recursively count usage only for those candidates
 4. stop counting once every candidate has exceeded the reportable threshold

This keeps the same behavior, but avoids a lot of type-checker and map work on declarations that are never going to produce diagnostics.

## Benchmark

Measured with a shallow clone of microsoft/vscode, using 6,187 src/**/*.ts(x) files through local tsgolint headless.

Environment:

- rule: no-unnecessary-type-parameters

Hyperfine, 3 runs:

| Binary | Mean | User | System |
| --- | ---: | ---: | ---: |
| baseline | 11.333s ± 0.061s | 43.635s | 0.992s |
| patched | 4.002s ± 0.078s | 14.775s | 0.570s |

That is a 2.83x speedup for this isolated rule workload.

fixes #961

### DIFF
--- a/internal/rules/no_unnecessary_type_parameters/no_unnecessary_type_parameters.go
+++ b/internal/rules/no_unnecessary_type_parameters/no_unnecessary_type_parameters.go
@@ -196,25 +196,35 @@ func classMembers(node *ast.Node) []*ast.Node {
 	}
 }
 
-func countTypeParameterUsage(ctx rule.RuleContext, node *ast.Node) map[*ast.Symbol]int {
+func countTypeParameterUsage(ctx rule.RuleContext, node *ast.Node, targetSymbols map[*ast.Symbol]struct{}) map[*ast.Symbol]int {
 	counts := make(map[*ast.Symbol]int)
+	remainingTargets := len(targetSymbols)
 
 	if ast.IsClassLike(node) {
 		for _, typeParameter := range node.TypeParameters() {
-			collectTypeParameterUsageCounts(ctx, typeParameter, counts, true)
+			if remainingTargets == 0 {
+				break
+			}
+			remainingTargets = collectTypeParameterUsageCounts(ctx, typeParameter, counts, targetSymbols, remainingTargets, true)
 		}
 		if heritageClauses := utils.GetHeritageClauses(node); heritageClauses != nil {
 			for _, heritageClause := range heritageClauses.Nodes {
 				for _, heritageType := range heritageClause.AsHeritageClause().Types.Nodes {
-					collectTypeParameterUsageCounts(ctx, heritageType, counts, true)
+					if remainingTargets == 0 {
+						break
+					}
+					remainingTargets = collectTypeParameterUsageCounts(ctx, heritageType, counts, targetSymbols, remainingTargets, true)
 				}
 			}
 		}
 		for _, member := range classMembers(node) {
-			collectTypeParameterUsageCounts(ctx, member, counts, true)
+			if remainingTargets == 0 {
+				break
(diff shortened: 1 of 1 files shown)


camc314:
### Merge activity

* **May 18, 7:11 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **May 18, 7:11 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **May 18, 7:18 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

connorshea:
Thank you!!

PR #977, opened 2026-05-19
test(no-floating-promises): re-enable recursive type argument cases

camc314:
Re-enable the valid and invalid recursive `CustomNode` type-argument cases. The handled async call must lint cleanly, while the unhandled call must still report a floating promise.

Both cases now pass with the repository's current typescript-go revision using normal call resolution, so no checker workaround is needed. The snapshot reflects the current diagnostic labels and ranges.

#978 is stacked on this PR.

### DIFF
--- a/internal/rule_tester/__snapshots__/no-floating-promises.snap
+++ b/internal/rule_tester/__snapshots__/no-floating-promises.snap
@@ -1684,3 +1684,38 @@ Help: The promise must end with a call to .catch, or end with a call to .then wi
   Suggestion 1: [floatingFixVoid] Add void operator to ignore.
   Suggestion 2: [floatingFixAwait] Add await operator.
 ---
+
+[TestNoFloatingPromisesRule/invalid-101 - 1]
+Diagnostic 1: floatingVoid (14:9 - 20:13)
+Message: Promises must be awaited, add void operator to ignore.
+Help: The promise must end with a call to .catch, or end with a call to .then with a rejection handler, or be explicitly marked as ignored with the `void` operator.
+  13 | 
+  14 |         (async () => {
+     |         ~~~~~~~~~~~~~~
+  15 |           wrapNode(() => {
+     | ~~~~~~~~~~~~~~~~~~~~~~~~~~
+  16 |             const node = createNode();
+     | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  17 | 
+  18 |             return wrapNode<typeof node.getNextNode<any>>(node.getNextNode);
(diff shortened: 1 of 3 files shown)


amirt01:
Hit this exact failure mode from a real-world library, not just a synthetic test case — wanted to add it as a data point since it shows the recursive-type pattern this PR guards against does occur in published `.d.ts` files.

Upgrading `three`/`@types/three` from `0.184.0` → `0.185.0` makes `tsgolint` (via `oxlint --type-aware`) hang indefinitely on this file (lints clean, 0 warnings/errors, on `0.184.0`):

```ts
import { float } from "three/tsl";
import type Node from "three/src/nodes/core/Node.js";

const x = float(1.0);

export async function check() {
  await Promise.resolve(x).then((val: Node) => {
    console.log(val);
  });
}
```

- `tsc --noEmit` on the same file finishes in well under a second with no errors on both versions.
- `oxlint --type-aware` finishes instantly on `0.184.0`, hangs indefinitely on `0.185.0` (killed after 60+ minutes of CPU-pegged runtime).
- Root cause on the types side: `@types/three@0.185.0` refactored `Node<TNodeType>` (`src/nodes/core/Node.d.ts`) to use unconstrained generic interfaces (`FloatOrVecExtensions<TNodeType>`, `IntOrVecExtensions<TNodeType>`, `UintOrVecExtensions<TNodeType>`, `BoolOrVecExtensions<TNodeType>`) merged via `declare module` across many files (`MathNode.d.ts`, `OperatorNode.d.ts`, etc.), where every merged method returns `Node<TNodeType>` — a self-reference back into the same conditional type, now fanned out across dozens of cross-file merges instead of one concrete interface per type. That's structurally the same "recursive type forces call-site instantiation" shape that this PR is narrowing the rule's fast path around.

Versions: `three@0.185.0`, `@types/three@0.185.0`, `oxlint@1.71.0`, `oxlint-tsgolint@0.23.0`.

Happy to share the full repro repo if useful — currently it's a 4-file project (`package.json`, `tsconfig.json`, `repro.ts`).

camc314:
### Merge activity

* **Sep 16, 7:58 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Sep 16, 7:58 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Sep 16, 8:04 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

PR #978, opened 2026-05-19
test(no-floating-promises): re-enable node test allowlist case

camc314:
Stacked on #977; merge #977 first.

Re-enable the `allowForKnownSafeCalls` case for `it` from `node:test`. Use a self-contained ambient declaration fixture with a promise-returning function and a negative control that reports the same call without the allowlist, so the test does not pass merely because Node's declarations are unavailable.


camc314:
Unneeded since #977 is now merged.

PR #983, opened 2026-05-23
fix(justfile): use --ignore-workspace for e2e pnpm commands

swandir:
#968 added PNPM workspace configuration and updated CI commands to use `--ignore-workspace`. The justfile still runs pnpm commands as before, causing pnpm to ignore the nested `e2e` package since it's not listed in the workspace. Because of that `just test` fails due to missing `vitest` on a fresh clone.


camc314 (approved):
thansk!

PR #992, opened 2026-05-29
Fix #990 no-unnecessary-type-assertion false negative

### LINKED ISSUE #990: `no-unnecessary-type-assertion` false negative on `'literal' as string | null`
# `no-unnecessary-type-assertion` false negative on `'literal' as string | null`

## Summary

`oxlint-tsgolint` does not report `@typescript-eslint/no-unnecessary-type-assertion` for an unnecessary assertion from a string literal to a nullable wider type.

```ts
const x: string | null = 'a' as string | null
```

The assertion is unnecessary because the original expression type is assignable to the receiver type. The official `typescript-eslint` implementation reports this case, but `oxlint-tsgolint` does not.

This appears to be a deterministic false negative in the rule implementation, not a sporadic monorepo/concurrency issue, because it is reproducible in a minimal single-file project.

## Environment

* `oxlint`: `1.64.0`
* `oxlint-tsgolint`: `0.22.1`
* Also reproduced with `oxlint-tsgolint`: `0.23.0`
* `typescript`: `5.9.3`
* `typescript-eslint`: `8.59.3` used for comparison
* Node.js: `24.x`
* OS: Linux, WSL2 / Ubuntu

## Minimal reproduction

```bash
mkdir tsgolint-repro && cd tsgolint-repro
npm init -y
npm install -D oxlint@1.64.0 oxlint-tsgolint@0.22.1 typescript@5.9.3

cat > .oxlintrc.json << 'EOF'
{
  "plugins": ["typescript"],
  "rules": {
    "@typescript-eslint/no-unnecessary-type-assertion": "error"
  },
  "options": { "typeAware": true }
}
EOF

cat > tsconfig.json << 'EOF'
{
  "compilerOptions": {
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "Bundler",
    "strict": true,
    "noEmit": true
  },
  "include": ["test.ts"]
}
EOF

cat > test.ts << 'EOF'
const x: string | null = 'a' as string | null
void x
EOF

npx oxlint --type-aware test.ts
```

## Actual behavior

`oxlint-tsgolint` reports no error.

```text
Found 0 warnings and 0 errors.
Finished in 119ms on 1 file with 85 rules using 8 threads.
```

Exit code: `0`.

## Expected behavior

`@typescript-eslint/no-unnecessary-type-assertion` should report the assertion on:

```ts
'a' as string | null
```

Reason:

* The expression `'a'` has the string literal type `'a'`.
* The type `'a'` is assignable to `string | null`.
* The receiver already has the type annotation `: string | null`.
* Therefore, the assertion does not make the assignment valid; it is redundant.

The code can be written without the assertion:

```ts
const x: string | null = 'a'
void x
```

## Comparison with `typescript-eslint`

Using the official `typescript-eslint` implementation on the same `test.ts`:

```bash
npm install -D eslint@9.30.0 typescript-eslint@8.59.3

cat > eslint.config.mjs << 'EOF'
import tseslint from 'typescript-eslint'

export default tseslint.config({
  files: ['test.ts'],
  extends: [tseslint.configs.recommendedTypeChecked],
  languageOptions: {
    parserOptions: {
      projectService: true,
      tsconfigRootDir: import.meta.dirname,
    },
  },
  rules: {
    '@typescript-eslint/no-unnecessary-type-assertion': 'error',
  },
})
EOF

npx eslint test.ts
```

Output:

```text
/tmp/tsgolint-repro/test.ts
  1:26  error  This assertion is unnecessary since the receiver accepts the original type of the expression  @typescript-eslint/no-unnecessary-type-assertion

✖ 1 problem (1 error, 0 warnings)
```

Exit code: `1`.

So the official `typescript-eslint` implementation reports the assertion as unnecessary, while `oxlint-tsgolint` does not.

## Why this matters

This false negative weakens migration from `typescript-eslint` typed linting to `oxlint-tsgolint`.

The pattern is common in test fixtures and ad-hoc casts, for example:

```ts
const value: string | null = 'fixture' as string | null
```

In this case, the receiver type already accepts the original expression, so the assertion should be removable.

## Notes

* This reproduces consistently in a minimal single-file project.
* This is different from sporadic monorepo/concurrency false positives because:

  * it is a false negative, not a false positive;
  * it does not require parallel execution;
  * it does not require a monorepo;
  * it reproduces deterministically.
* The same behavior was also confirmed with `oxlint-tsgolint@0.23.0`.
* This may be related to receiver-type checking for assertions where the asserted type is a nullable union wider than the source expression type.

lapinChiro:


camc314:
Thanks for putting this PR up.

I've fixed it in #1001 while fixing a large number of other similar issues - thanks!

PR #996, opened 2026-06-01
fix(prefer-optional-chain): suggest unsafe nullish rewrite

### LINKED ISSUE #22860: linter: type-aware prefer-optional-chain --fix unsafely rewrites "x === undefined || x.prop === null" (changes behavior)

oxlint 1.67.0, oxlint-tsgolint 0.23.0 (also reproduces with `latest`, which currently resolves to these).

### What command did you run?

```
oxlint --fix repro.ts
```

```jsonc
{
  "plugins": ["typescript"],
  "options": { "typeAware": true, "typeCheck": true },
  "rules": { "typescript/prefer-optional-chain": "error" }
}
```

With a standard strict `tsconfig.json`:

```jsonc
{
  "compilerOptions": {
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "bundler",
    "strict": true,
    "noEmit": true
  },
  "include": ["*.ts"]
}
```

### What happened?

Under **type-aware** linting, the `typescript/prefer-optional-chain` autofix applied by plain `--fix` (the *safe* tier — `--fix-suggestions` is the one documented as "may change program behavior") rewrites a strict-equality `||` guard into an optional chain that **changes runtime behavior**. The rewritten output still type-checks, so the change is silent.

**Minimal reproduction** (`repro.ts`):

```ts
interface Box {
    value: string | null;
}

// Returns true when there is no first item OR its value is null.
export function isBlocked(items: Box[]): boolean {
    const item = items.at(0);
    if (item === undefined || item.value === null) {
        return true;
    }
    return false;
}
```

Running `oxlint --fix repro.ts` rewrites the guard to:

```ts
    if (item?.value === null) {
```

**Behavior change:**

| call | before fix | after fix |
| --- | --- | --- |
| `isBlocked([])` | `true` — `item` is `undefined`, so `item === undefined` is `true` | `false` — `item?.value` is `undefined`, and `undefined === null` is `false` |

So `isBlocked([])` flips from `true` to `false`. `tsc --noEmit` still passes on the fixed output, so nothing surfaces the regression.

**Why it is unsound:** collapsing to `item?.value === null` is only equivalent for a **loose** `== null` comparison (since `undefined == null` is `true`). With strict `=== null` and the separate `item === undefined` guard, the "item is undefined" branch is silently dropped.

This appears to share a root cause with #20054 (closed) — `undefined`/`null` strict-equality truthiness making the collapse unsound — but it is a different, still-unfixed syntactic pattern (`||` with strict `=== undefined` / `=== null`, rather than the `&&` ternary in that report). It still reproduces on the latest release.

**Expected:** plain `--fix` should not apply this rewrite because it changes behavior (at most surface it under `--fix-suggestions`); ideally the rule should not propose the collapse under strict equality where it is not semantics-preserving.

camc314:
**Summary**

- Treat optional-chain rewrites like `item === undefined || item.value === null` as suggestions instead of autofixes.
- Add a regression case and snapshot for the unsafe strict undefined-prefix plus strict null comparison pattern.
- Keep the new structural check lazy so it only runs when the diagnostic path still needs to decide between a fix and a suggestion.

Fixes oxc-project/oxc#22860

### DIFF
--- a/internal/rule_tester/__snapshots__/prefer-optional-chain.snap
+++ b/internal/rule_tester/__snapshots__/prefer-optional-chain.snap
@@ -5379,3 +5379,13 @@ Message: Prefer using an optional chain expression instead, as it's more concise
      |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    5 | 
 ---
+
+[TestPreferOptionalChainRule/invalid-600 - 1]
+Diagnostic 1: preferOptionalChain (3:1 - 3:41)
+Message: Prefer using an optional chain expression instead, as it's more concise and easier to read.
+   2 | declare const item: { value: string | null } | undefined;
+   3 | item === undefined || item.value === null;
+     | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+   4 | 
+  Suggestion 1: [optionalChainSuggest] Change to an optional chain.
+---
--- a/internal/rules/prefer_optional_chain/prefer_optional_chain.go
+++ b/internal/rules/prefer_optional_chain/prefer_optional_chain.go
@@ -103,6 +103,32 @@ func isStrictNullComparison(op Operand) bool {
 	return isComparisonAgainst(op, utils.IsNullLiteral)
 }
 
+func isStrictNullCheckForOrChain(op Operand) bool {
+	switch op.typ {
+	case OperandTypeStrictEqualNull, OperandTypeNotStrictEqualNull:
+		return true
+	case OperandTypeComparison:
+		if bin := unwrapBinary(op.node); bin != nil {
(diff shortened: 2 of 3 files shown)


camc314:
### Merge activity

* **Jun 1, 7:47 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Jun 1, 7:47 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Jun 1, 7:55 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

PR #1004, opened 2026-06-03
fix(no-unnecessary-type-arguments): preserve constructor inference

### LINKED ISSUE #22915: linter: typescript/no-unnecessary-type-arguments autofix removes default generic required for constructor inference

1.68.0

### What command did you run?

pnpm exec oxlint --fix repro.ts

```ts
import { defineConfig } from 'oxlint';

export default defineConfig({
  options: {
    typeAware: true,
  },
  rules: {
    'typescript/no-unnecessary-type-arguments': 'error',
  },
});
```

### What happened?

`typescript/no-unnecessary-type-arguments` reports and autofixes a default generic type argument on a constructor call:

```ts
const context = new RequestContext<unknown>([
  ['instructions', 'Reply for the agent.'],
  ['tools', {}],
]);
```
The fixer changes it to:

```ts
const context = new RequestContext([
  ['instructions', 'Reply for the agent.'],
  ['tools', {}],
]);
```
After the fix, TypeScript fails because the constructor infers a narrower RequestContext<{ ... }> instead of the intended untyped/default RequestContext<unknown>.

### Minimal repro
```ts
type RecordToTuple<T> = {
  [K in keyof T]: [K, T[K]];
}[keyof T][];

class RequestContext<
  Values extends Record<string, unknown> | unknown = unknown,
> {
  constructor(
    _iterable?: Values extends Record<string, unknown>
      ? RecordToTuple<Partial<Values>>
      : Iterable<readonly [string, unknown]>,
  ) {}

  set<K extends Values extends Record<string, unknown> ? keyof Values : string>(
    _key: K,
    _value: Values extends Record<string, unknown>
      ? K extends keyof Values
        ? Values[K]
        : never
      : unknown,
  ): void {}
}

type UntypedRequestContext = RequestContext;

function acceptsUntypedContext(_context: UntypedRequestContext): void {}

const context = new RequestContext<unknown>([
  ['instructions', 'Reply for the agent.'],
  ['tools', {}],
]);

acceptsUntypedContext(context);
```

### Expected behavior
The rule should not report/fix this type argument, because removing it changes the inferred constructor type and breaks downstream assignability.

`tsc --noEmit` passes before the fix.

### Actual behavior
Oxlint removes `<unknown>`.

Then `tsc --noEmit` fails:
```
error TS2345: Argument of type 'RequestContext<{ [x: string]: {}; } & ...>'
is not assignable to parameter of type 'UntypedRequestContext'.
  Type 'unknown' is not assignable to type '{ [x: string]: {}; } & ...'
```

### Notes
This appears related to the remaining default-type-argument path after inference reporting was reverted in oxc-project/tsgolint#892.

In this case the explicit type argument equals the default, but omitting it does not preserve semantics because constructor inference chooses a narrower type from the argument list.

camc314:
`typescript/no-unnecessary-type-arguments` could remove an explicit default class type argument from `new` expressions even when constructor arguments would infer a different class type after the type argument list is removed. In the reported `RequestContext<unknown>(...)` case, the explicit `unknown` matches the default but is still semantically meaningful because omitting it narrows the constructed type from the iterable argument.

This change keeps reporting default type arguments for constructors when they are safe, but skips the fix when constructor parameter type annotations reference class type parameters that would become inferable from constructor arguments. A regression test covers the issue sample.

Fixes oxc-project/oxc#22915

### DIFF
--- a/internal/rule_tester/__snapshots__/no-unnecessary-type-arguments.snap
+++ b/internal/rule_tester/__snapshots__/no-unnecessary-type-arguments.snap
@@ -250,3 +250,12 @@ Message: This is the default value for this type parameter, so it can be omitted
      |   ~~~~~
    6 |       
 ---
+
+[TestNoUnnecessaryTypeArguments/invalid-28 - 1]
+Diagnostic 1: unnecessaryTypeParameter (5:15 - 5:21)
+Message: This is the default value for this type parameter, so it can be omitted.
+   4 | }
+   5 | new C<string, unknown>('value');
+     |               ~~~~~~~
+   6 |       
+---
--- a/internal/rules/no_unnecessary_type_arguments/no_unnecessary_type_arguments.go
+++ b/internal/rules/no_unnecessary_type_arguments/no_unnecessary_type_arguments.go
@@ -120,7 +120,67 @@ var NoUnnecessaryTypeArgumentsRule = rule.Rule{
 			return nil
 		}
 
-		checkArgsAndParameters := func(arguments *ast.NodeList, parameters []*ast.Node) {
+		typeNodeReferencesTypeParameter := func(typeNode *ast.Node, typeParameterSymbols map[*ast.Symbol]struct{}) bool {
+			var visit func(node *ast.Node) bool
+			visit = func(node *ast.Node) bool {
+				if node == nil {
+					return false
+				}
+
+				if ast.IsIdentifier(node) {
+					if symbol := ctx.TypeChecker.GetSymbolAtLocation(node); symbol != nil {
+						if _, ok := typeParameterSymbols[symbol]; ok {
+							return true
+						}
+					}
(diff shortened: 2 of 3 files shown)


camc314:
### Merge activity

* **Jun 3, 10:46 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Jun 3, 10:51 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Jun 3, 10:59 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

PR #1028, opened 2026-06-25
fix(dot-notation): determine the relevant accessor

### LINKED ISSUE #1027: `typescript/dot-notation` ignores `allowPrivate`/`ProtectedClassPropertyAccess` for accessors with a non-public setter
## Versions

oxlint 1.71.0 · oxlint-tsgolint 0.23.0

## Setup

`.oxlintrc.json`

```json
{
	"rules": {
		"typescript/dot-notation": ["error", {
			"allowPrivateClassPropertyAccess": true,
			"allowProtectedClassPropertyAccess": true
		}]
	}
}
```

`tsconfig.json`

```json
{
	"compilerOptions": { "strict": true },
	"include": ["index.ts"]
}
```

`package.json`

```json
{
	"scripts": { "fix": "oxlint --type-aware --fix" },
	"devDependencies": { "oxlint": "1.71.0", "oxlint-tsgolint": "0.23.0" }
}
```

## Input — `index.ts`

```ts
class Example {
	private _a = 0;
	public get prot(): number {
		return this._a;
	}
	protected set prot(v: number) {
		this._a = v;
	}

	private _b = 0;
	public get priv(): number {
		return this._b;
	}
	private set priv(v: number) {
		this._b = v;
	}
}

declare const obj: Example;

obj["prot"] = 1;
obj["priv"] = 1;
```

## Expected

No diagnostics — both setters are non-public and both `allow*` options are `true`.

## Actual

```
index.ts:21:5: error typescript(dot-notation): ["prot"] is better written in dot notation.
index.ts:22:5: error typescript(dot-notation): ["priv"] is better written in dot notation.
```

`--fix` rewrites both to dot notation (`obj.prot = 1;`, `obj.priv = 1;`), neither of which compiles.
Visibility appears to be read from the (public) getter, ignoring the setter actually used in the write.

AlCalzone:
fixes: #1027 

The rule simply used the first accessor declaration to determine the visibility of a property. This breaks down when the getter and setter have different visibilities. With this PR we try to determine the actual accessor being used and prefer its visibility for the check.


camc314 (approved):
thanks!

PR #1031, opened 2026-06-28
perf(consistent-return): defer per-function type resolution

connorshea:
Generated with Claude Code, reviewed and tested by me.

`consistent-return` previously did eager, per-function work on **every** function it entered, even though most functions never trigger a diagnostic:

- `isReturnVoidOrThenableVoid(ctx, node)` — a `GetTypeAtLocation` call — was run for every function on `enterFunction`.
- `getFunctionNameWithKind(...)` was computed up front and stored on every `functionState`, even when no report is ever emitted.

This PR defers both:

- **`allowsVoidReturn` is now lazy + memoized.** The `GetTypeAtLocation` call only happens when a `return;` statement or the function-exit check actually needs it. Functions that never reach those branches pay nothing.
- **The function name is computed only at report time**, where it is actually used, instead of for every function entered.

The visitor logic is otherwise unchanged; the exit-check condition is just reordered so the cheap flag checks short-circuit before the (now lazy) `allowsVoidReturn` call.

## Performance

Benchmarked over the [VS Code](https://github.com/microsoft/vscode) repo — 11,674 files, 18 threads, **646,186 rule calls**, 12 runs per binary, per-rule timing via `oxlint --debug=timings`.

| Metric | before | after | improvement |
|---|---:|---:|---:|
| mean | 7315.8 ms | 854.9 ms | **8.56× faster** |
| median | 7192.0 ms | 843.7 ms | 8.52× faster |
| min | 6755.7 ms | 782.5 ms | 8.63× faster |
| max | 8649.8 ms | 981.6 ms | 8.81× faster |
| stddev | 508.7 ms | 55.5 ms | — |

## Correctness

Diagnostics are unchanged. Running the rule in isolation over the same repo produces byte-for-byte identical output before and after (34,596 lines incl. warnings, sorted & diffed): **2035 warnings + 48 errors**, identical.

## Methodology

```bash
OXLINT_TSGOLINT_PATH=<binary> oxlint \
  --type-aware --quiet -A all -W typescript/consistent-return \
  --debug=timings -f default
```

### DIFF
--- a/internal/rules/consistent_return/consistent_return.go
+++ b/internal/rules/consistent_return/consistent_return.go
@@ -25,14 +25,14 @@ func buildUnexpectedReturnValueMessage(functionNameWithKind string) rule.RuleMes
 }
 
 type functionState struct {
-	node                 *ast.Node
-	upper                *functionState
-	hasReturn            bool
-	hasReturnValue       bool
-	hasMismatch          bool
-	allowsVoidReturn     bool
-	functionNameWithKind string
-	messageId            string
+	node                     *ast.Node
+	upper                    *functionState
+	hasReturn                bool
+	hasReturnValue           bool
+	hasMismatch              bool
+	allowsVoidReturn         bool
+	allowsVoidReturnComputed bool
+	messageId                string
 }
 
 func getFunctionNameWithKind(sourceFile *ast.SourceFile, node *ast.Node) string {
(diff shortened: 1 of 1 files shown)


camchenry (approved):
Makes sense and seems safe given we're doing the same work just lazily.

camc314 (approved):

PR #1033, opened 2026-06-28
perf(strict-void-return): reuse function-body walk visitor closure

connorshea:
## Summary

`strict-void-return` walks each function body (`visit`) looking for non-void
`return` statements. The `ForEachChild` callback was declared **inline inside
the recursive `visit`**, so Go heap-allocated a fresh closure on **every
recursive call — i.e. once per node in the body**.

Hoisting that callback to a single `visitChild` closure (allocated once and
reused for the whole walk) removes the per-node allocation. Behavior is
unchanged — the callback body is identical, it's just no longer re-created at
every node.

```go
var visit func(node *ast.Node)
// allocated once, reused for every node
visitChild := func(child *ast.Node) bool {
    visit(child)
    return false
}
visit = func(node *ast.Node) {
    ...
    node.ForEachChild(visitChild) // was: node.ForEachChild(func(child *ast.Node) bool { visit(child); return false })
}
```

## Benchmark

Synthetic workload: 3,000 void-context callbacks (`arr.forEach((x) => { … })`),
each with ~20 statements of nested expressions plus a non-void `return` (so the
rule does not early-out and actually walks the body). Allocations measured with
Go's heap profiler at `-memprofilerate=1` (every allocation sampled, so the
counts are exact). The generated source is byte-identical before and after, so
the entire delta is attributable to this change.

| Metric          | Before (inline closure) | After (hoisted) | Change            |
| --------------- | ----------------------: | --------------: | ----------------- |
| **alloc_objects** |             4,439,970 |         512,916 | **−3.93M (−88%)** |
| alloc_space     |                543.6 MB |        483.2 MB | −60 MB (−11%)     |
| diagnostics     |                   3,000 |           3,000 | unchanged         |

The per-node closure was being allocated for every node in every body walk
(~3.9M closures across the workload), all of which are eliminated. Each closure
is tiny, so the byte reduction is smaller (~11%), but GC cost is driven largely
by **object count** — cutting 88% of allocations is a direct reduction in GC
pressure on large/closure-heavy files.

## Notes

- Behavior is unchanged; the rule's existing tests pass and produce an identical
  number of diagnostics.
- Same pattern (inline `ForEachChild`/`ForEachChildAndJSDoc` callback inside a
  recursive walk) exists in a couple of other rules; this PR covers
  `strict-void-return` only, which walks whole function bodies and so benefits
  the most.

### DIFF
--- a/internal/rules/strict_void_return/strict_void_return.go
+++ b/internal/rules/strict_void_return/strict_void_return.go
@@ -111,9 +111,6 @@ var StrictVoidReturnRule = rule.Rule{
 			}
 
 			var visit func(node *ast.Node)
-			// Allocate the ForEachChild visitor once and reuse it for every node.
-			// Declaring the callback inline inside visit would heap-allocate a new
-			// closure on every recursive call (once per node in the function body).
 			visitChild := func(child *ast.Node) bool {
 				visit(child)
 				return false
--- a/internal/rules/strict_void_return/strict_void_return.go
+++ b/internal/rules/strict_void_return/strict_void_return.go
@@ -111,6 +111,13 @@ var StrictVoidReturnRule = rule.Rule{
 			}
 
 			var visit func(node *ast.Node)
+			// Allocate the ForEachChild visitor once and reuse it for every node.
+			// Declaring the callback inline inside visit would heap-allocate a new
+			// closure on every recursive call (once per node in the function body).
+			visitChild := func(child *ast.Node) bool {
+				visit(child)
+				return false
+			}
 			visit = func(node *ast.Node) {
 				if node == nil {
 					return
@@ -129,10 +136,7 @@ var StrictVoidReturnRule = rule.Rule{
 					}
 				}
 
-				node.ForEachChild(func(child *ast.Node) bool {
-					visit(child)
-					return false
-				})
+				node.ForEachChild(visitChild)
 			}
 			visit(funcNode)
 		}


jakebailey:
FWIW I also have https://github.com/microsoft/typescript-go/pull/4395 which will make `ForEachChild` not escape its func.

connorshea:
> FWIW I also have [microsoft/typescript-go#4395](https://github.com/microsoft/typescript-go/pull/4395) which will make `ForEachChild` not escape its func.

Thanks for calling that out! Makes sense to me, if we think that's likely to land upstream then I'm fine closing this PR and the other I just opened, since they'll be unnecessary once that change is in typescript-go.

jakebailey:
I think your PRs are fine; we have a lot of code in TS that does this same thing.

internal/rules/strict_void_return/strict_void_return.go
```diff
@@ -111,6 +111,13 @@ var StrictVoidReturnRule = rule.Rule{
 			}
 
 			var visit func(node *ast.Node)
+			// Allocate the ForEachChild visitor once and reuse it for every node.
+			// Declaring the callback inline inside visit would heap-allocate a new
+			// closure on every recursive call (once per node in the function body).
```
camc314: comment is unnecessary


```suggestion
```

camc314 (changes requested):
🚀

camc314 (approved):
:rocket:

PR #1039, opened 2026-06-29
refactor(no-unnecessary-condition): remove dead containsUnguardedElementAccess

connorshea:
This code was only referenced by itself, so it is dead code.

The live skip logic uses `hasUnsafeElementAccess`, which appears to have superseded it.


camc314 (approved):
nice! It's disappointing the linter doesn't catch this

PR #1059, opened 2026-07-07
fix(no-deprecated): avoid contextual union property false positives

### LINKED ISSUE #1046: false positive with `link` no-deprecated
Image

Any `link` property in plain object which passes to another function is now incorrectly reported with `typescript(no-deprecated): `link` is deprecated. A legacy feature for browser compatibility`.

A sample code below:

```ts
import { sidebar } from "vuepress-theme-hope";

export const enSidebar = sidebar({
  "/": [
    "",
    {
      text: "Demo",
      icon: "laptop-code",
      prefix: "demo/",
      link: "demo/",
      children: "structure",
    },
    {
      text: "Articles",
      icon: "book",
      prefix: "posts/",
      children: "structure",
    },
    "intro",
    "slide",
  ],
});
```

camc314:
## Summary

Fixes #1046.

`typescript/no-deprecated` no longer reports ordinary object-literal properties such as `link` merely because a contextual union also contains a primitive `string` constituent whose boxed `String` type has a deprecated member with the same name.

## Problem

The contextual object-literal property handling added for `no-deprecated` performed these steps:

1. Ask the checker for the object literal's contextual type with `getContextualType`.
2. Look up the object property's name on that type with `getPropertyOfType`.
3. Treat the property as deprecated when any declaration attached to the returned symbol has an `@deprecated` tag.

That is unsafe for contextual unions.

The VuePress `sidebar` helper in the issue accepts a recursive union whose item type is structurally equivalent to:

```ts
type SidebarItemOptions =
  | AutoLinkOptions
  | SidebarGroupOptions
  | SidebarStructureOptions
  | string;
```

For the nested object in the report, the checker produces an apparent contextual union containing object alternatives and the boxed `String` alternative. Looking up `link` on that union creates a synthetic property symbol. Its declarations include the normal, non-deprecated `link` properties from the sidebar object types as well as the legacy `String.prototype.link` declaration from `lib.es2015.core.d.ts`, which is marked deprecated for browser compatibility.

The rule's declaration scan used "any declaration is deprecated" semantics. As a result, the unrelated legacy `String.link` declaration contaminated the synthetic union property and produced:

```text
`link` is deprecated. A legacy feature for browser compatibility
```

## Fix

This change makes contextual property analysis follow the TypeScript checker's own semantics in two places:

- It uses `getApparentTypeOfContextualType` instead of raw `getContextualType`. The apparent-type helper instantiates contextual types and discriminates contextual unions using object-literal members before member lookup. The checker source explicitly distinguishes this from `getContextualType`, which is intended for propagating a contextual type rather than inspecting its members.
- It asks `isDeprecatedSymbol` whether the looked-up property symbol is deprecated before extracting the JSDoc reason. For synthetic union symbols, TypeScript requires all contributing declarations to be deprecated. This prevents a single deprecated member from an unrelated union constituent from marking the combined property deprecated, while preserving TypeScript's special handling for symbols such as merged interface members.

Both internal checker methods are exposed through the existing generated checker shim configuration, and the shim was regenerated with `just shim`.

## Regression test

The new valid case reproduces the relevant VuePress sidebar type shape:

- recursive sidebar item union;
- object and `string` constituents;
- nested object passed through a helper function;
- ordinary `link` property;
- `children: "structure"` discriminant.

Before the fix it reports the deprecated browser-compatibility `String.link`; after the fix it produces no diagnostic.


camc314:
### Merge activity

* **Jul 7, 10:17 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Jul 7, 10:31 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Jul 7, 10:38 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

PR #1061, opened 2026-07-09
fix(prefer-optional-chain): preserve negated optional guard

### LINKED ISSUE #1060: prefer-optional-chain --fix drops the outer negation on `x == null || !x.y`, inverting runtime behaviour
## Summary

`oxlint --type-aware --type-check --fix` for `typescript/prefer-optional-chain` inverts a boolean guard of the form `x == null || !x.y`. It rewrites the condition to `x?.y`, dropping the outer negation and inverting the runtime behaviour.

The diagnostic itself is correct (the expression *is* collapsible) but the autofix produces the negation of the original.

typescript-eslint's `prefer-optional-chain` (the rule this ports) fixes the same input to `!x?.y`, which preserves the semantics, so this is a divergence from the upstream fixer.

The practical impact is serious: `--fix` changes runtime behaviour in a way that still satisfies the compiler.

## Versions

- oxlint: 1.72.0
- oxlint-tsgolint: 0.24.0
- typescript: 5.9.3 (clean)
- reference: @typescript-eslint/eslint-plugin 8.46.3 (produces the correct fix — see below)
- Platform: macOS (darwin arm64)

## Minimal reproduction

`.oxlintrc.json`:

```jsonc
{ "rules": { "typescript/prefer-optional-chain": "error" } }
```

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "strict": true,
    "target": "ES2022",
    "module": "ESNext",
    "moduleResolution": "Bundler"
  },
  "include": ["repro.ts"]
}

```

`repro.ts`:

```ts
export function guard(item: { myField?: string } | undefined) {
  if (item == null || !item.myField) {
    throw new Error('missing');
  }
  return item.myField;
}
```

```console
$ npx oxlint --type-aware --type-check --fix repro.ts
$ cat repro.ts
export function guard(item: { myField?: string } | undefined) {
  if (item?.myField) {
    throw new Error('missing');
  }
  return item.myField;
}
```

Expected: `if (!item?.myField)`
Actual: `if (item?.myField)`

## Additional notes

- `!x || !x.y` (every operand negated) fixes correctly to `!x?.y`.
- `!x.y || x == null` (negation first) is not fixed at all. The trigger is a *leading* nullish-comparison operand followed by a negated member access 
- Not specific to loose equality `==`; `x === null || !x.y` and `x === undefined || !x.y` invert the same way.

camc314:
## Summary
- preserve the outer negation when OR-chain optional-chain fixes end in a negated property access
- add regression coverage for nullish guards followed by negated member access
- update the prefer-optional-chain snapshot

Fixes #1060.

### DIFF
--- a/internal/rule_tester/__snapshots__/prefer-optional-chain.snap
+++ b/internal/rule_tester/__snapshots__/prefer-optional-chain.snap
@@ -5396,3 +5396,30 @@ Message: Prefer using an optional chain expression instead, as it's more concise
    1 | foo && foo[bar as string | number] && foo[bar as string | number].baz;
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 ---
+
+[TestPreferOptionalChainRule/invalid-602 - 1]
+Diagnostic 1: preferOptionalChain (3:5 - 3:33)
+Message: Prefer using an optional chain expression instead, as it's more concise and easier to read.
+   2 | declare const item: { myField?: string } | undefined;
+   3 | if (item == null || !item.myField) {
+     |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+   4 |   throw new Error('missing');
+---
+
+[TestPreferOptionalChainRule/invalid-603 - 1]
+Diagnostic 1: preferOptionalChain (3:5 - 3:34)
+Message: Prefer using an optional chain expression instead, as it's more concise and easier to read.
+   2 | declare const item: { myField?: string } | null;
+   3 | if (item === null || !item.myField) {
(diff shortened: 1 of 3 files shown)


camc314:
### Merge activity

* **Jul 9, 9:00 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Jul 9, 9:00 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Jul 9, 9:09 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

PR #1065, opened 2026-07-10
perf(no-misused-promises): skip type queries for never-callable expressions

connorshea:
## Summary

Every `voidReturn` report ultimately requires `returnsThenable(expr)` — the expression's apparent type must have a call signature returning a thenable. Literals, template strings, array literals, and spread-free object literals can never have call signatures, so the type queries gating those reports can be skipped syntactically.

Object literals **with spreads** are still checked: spreading a generic value keeps its type (`{ ...t }` where `T extends () => Promise<void>` has type `T`), whose apparent type can have call signatures. `{ ...anyValue }` remains safe in principle (`any` has no call signatures), but the guard conservatively checks any literal containing a spread. This was a real false negative found in adversarial review (same latent bug as strict-void-return, fixed there in #1068) — regression tests cover both guarded paths (call argument and contextually typed variable initializer).

This adds a `couldBeFunctionLikeValue` kind check in front of the checker calls in:

- `voidFunctionArguments` — per-argument guard before `GetTypeOfSymbolAtLocation` on the parameter (positional, rest, and both rest array/tuple unwrap loops)
- `checkArguments` — pre-scan of all arguments before the callee signature resolution (also skips zero-arg calls, which previously paid the full callee + params walk)
- `checkProperty` — before `Checker_getContextualType` on property assignments
- `checkReturnStatement` / `checkAssignment` / `checkVariableDeclaration` — before the contextual/declared type queries

## Measurement

VS Code corpus (`src/tsconfig.json`, 7.1k files), `--debug timings`, all rules enabled, mean of 3 runs (at the head including the spread fix):

| | rule time (mean of 3) | calls |
|---|---|---|
| before | 4658 ms (4758 / 4553 / 4663) | 2,815,527 |
| after | 3804 ms (4025 / 3730 / 3656) | 2,815,527 |

Diagnostic parity is exact: 207,794 diagnostics compared as `(rule, file, pos, end, messageId)` keys, 0 differing keys. As with #1064, the win is partially masked in an all-rules run because other type-aware rules warm the checker's caches on the same nodes; in subset configs the avoided queries are paid in full.

Known theoretical limitation (shared with #1068): global-interface augmentation (e.g. `interface String { (): Promise<void> }`) can give literal types call signatures, which the guard would skip. This is pathological declaration merging and considered acceptable.

Rule test suite passes, including two new regression tests for the spread case.

### DIFF
--- a/internal/rule_tester/__snapshots__/no-misused-promises.snap
+++ b/internal/rule_tester/__snapshots__/no-misused-promises.snap
@@ -1007,3 +1007,21 @@ Message: Promise-returning function provided to property where a void return was
      |                ~~~~~~~~~~~~~~~
    5 | };
 ---
+
+[TestNoMisusedPromisesRule/invalid-90 - 1]
+Diagnostic 1: voidReturnArgument (4:11 - 4:18)
+Message: Promise returned in function argument where a void return was expected.
+   3 | function f<T extends () => Promise<void>>(t: T) {
+   4 |   takesCb({ ...t });
+     |           ~~~~~~~~
+   5 | }
+---
+
+[TestNoMisusedPromisesRule/invalid-91 - 1]
+Diagnostic 1: voidReturnVariable (3:26 - 3:33)
+Message: Promise-returning function provided to variable where a void return was expected.
+   2 | function f<T extends () => Promise<void>>(t: T) {
+   3 |   const cb: () => void = { ...t };
+     |                          ~~~~~~~~
+   4 | }
+---
--- a/internal/rules/no_misused_promises/no_misused_promises.go
+++ b/internal/rules/no_misused_promises/no_misused_promises.go
@@ -117,6 +117,30 @@ var NoMisusedPromisesRule = rule.Rule{
 			checksVoidReturnOpts = defaultChecksVoidReturnOpts()
 		}
 
+		// A syntactic check for whether an expression could possibly evaluate to a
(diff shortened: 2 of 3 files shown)


camc314:
> Known theoretical limitation (shared with https://github.com/oxc-project/tsgolint/pull/1068): global-interface augmentation (e.g. interface String { (): Promise<void> }) can give literal types call signatures, which the guard would skip. This is pathological declaration merging and considered acceptable.

I missed this on #1068 (i'm going to revert) but we should be targeting correctness over performance i these cases. If it was the reverse (excess diagnostics rather than missed), it would be slightly more acceptable.

connorshea:
> > Known theoretical limitation (shared with #1068): global-interface augmentation (e.g. interface String { (): Promise }) can give literal types call signatures, which the guard would skip. This is pathological declaration merging and considered acceptable.
> 
> I missed this on #1068 (i'm going to revert) but we should be targeting correctness over performance i these cases. If it was the reverse (excess diagnostics rather than missed), it would be slightly more acceptable.

fair! Sorry for not opting to skip these two PRs on that basis.

camc314:
> fair! Sorry for not opting to skip these two PRs on that basis.

no worries!

connorshea:
I'm pretty sure we can try this again but with the edge case handled. I don't know if it'll have any performance benefit at that point, but I'll try it later and bench it and see.

camc314:
sounds good!

PR #1074, opened 2026-07-10
perf(no-unnecessary-type-arguments): gate type queries behind cheap checks

macalinao:
## Summary

`no-unnecessary-type-arguments` ran its expensive type-parameter queries unconditionally. On a real React codebase this made it the **single most expensive rule** — ~15–18% of total lint time — because React's deeply-generic hook overloads (`useState`, `useRef`, …) are pathological for signature resolution. Three complementary short-circuits, all producing **identical diagnostics**:

1. **Skip the query for nodes with no explicit type argument list** (all 7 listeners). A node can only be reported if it has a `<…>` list, so this moves `checkArgsAndParameters`'s existing empty-arguments early-return *ahead* of the eagerly-evaluated query argument.

2. **A cheap symbol-based gate for the call/new/tagged path.** The rule can only report when the *matched* overload's last type parameter has a **default**. Instead of always paying `getResolvedSignature` (full overload resolution + inference), resolve the callee **symbol** (a cheap name lookup) and skip resolution when no overload that could accept this call's *argument count* carries a defaulted type parameter. React's `useState<S = undefined>()` default lives on the **zero-argument** overload, so `useState<string>("")` is inapplicable to it by arity and resolution selects the argument-taking overload, which has no default. (Verified with `tsc`: with the zero-arg defaulted overload declared *first*, `f<string>("")` still resolves to the arg-taking overload.) Any uncertainty (unresolved symbol, class/variable declaration, tagged-template/JSX where arg count is ambiguous) falls back to resolving. **False-negative-free**: the overload that would report always accepted the call's arguments, so it is never excluded and always forces resolution.

3. **Skip cloning + sorting `symbol.Declarations` for single-declaration symbols** in the type-reference path. The clone+sort only matters for merged symbols (multiple declarations); the common case is one declaration, where it just allocates.

## Measurement — real React codebase (excalidraw)

`--debug timings`, all rules enabled, `packages/excalidraw/tsconfig.json` (with `@types/react` installed). The row is this rule's own listener time:

| | rule time (3 runs) | % of total | calls |
|---|---|---|---|
| **before** | 1571 / 1336 / 1156 ms | 15–18% | 20,277 |
| **after** | 11 / 15 / 21 ms | 0.1–0.3% | 20,277 |

**~99% reduction** (~1.3 s → ~15 ms), identical call count. Diagnostics **byte-identical** — this rule's 4 findings match exactly, and all 7,458 diagnostics across every rule are unchanged (zero collateral).

### Microbenchmarks (single-rule isolation, AMD Ryzen 9 9950X)

| Scenario | Before | After | Δ |
|---|---|---|---|
| typeorm `src` (507 files), warm (benchstat n=6) | 30.39ms | 19.23ms | **−36.7%** (p=0.002) |
| allocations (after clone-skip) | 20.14k | 9.7k | **−52%** |
| React `useState/useRef/useMemo<T>` cold, 6000 calls | 47ms | 20ms | **−57%** (near the no-op floor) |

## Reproduce on excalidraw

```sh
# 1. Corpus with real @types/react (the generic hook overloads)
git clone --depth 1 https://github.com/excalidraw/excalidraw
( cd excalidraw && yarn install --frozen-lockfile )

# 2. Build the "after" binary from this branch
go build -o /tmp/tsgolint-after ./cmd/tsgolint
# ...and a "before" binary from the base commit (stash this rule's changes or check out main)

# 3. Compare this rule's own time (all rules, --debug timings)
cd excalidraw
/tmp/tsgolint-before --tsconfig packages/excalidraw/tsconfig.json --debug timings | grep no-unnecessary-type-arguments
/tmp/tsgolint-after  --tsconfig packages/excalidraw/tsconfig.json --debug timings | grep no-unnecessary-type-arguments

# 4. Verify diagnostics are unchanged
diff \
  <(/tmp/tsgolint-before --tsconfig packages/excalidraw/tsconfig.json 2>/dev/null | sed 's/\x1b\[[0-9;]*m//g') \
  <(/tmp/tsgolint-after  --tsconfig packages/excalidraw/tsconfig.json 2>/dev/null | sed 's/\x1b\[[0-9;]*m//g')
```

## Tests

Rule suite passes. Adds two regression tests:
- **valid** `useThing<string>('')` — defaulted type parameter only on the zero-arg overload, must not false-report (the `useState` shape).
- **invalid** `make<number>(1)` — the argument-taking overload itself carries the default, so the gate must resolve and still report the redundant type argument (snapshot `invalid-29`).

### DIFF
--- a/internal/rule_tester/__snapshots__/no-unnecessary-type-arguments.snap
+++ b/internal/rule_tester/__snapshots__/no-unnecessary-type-arguments.snap
@@ -259,3 +259,12 @@ Message: This is the default value for this type parameter, so it can be omitted
      |               ~~~~~~~
    6 |       
 ---
+
+[TestNoUnnecessaryTypeArguments/invalid-29 - 1]
+Diagnostic 1: unnecessaryTypeParameter (4:6 - 4:11)
+Message: This is the default value for this type parameter, so it can be omitted.
+   3 | declare function make(): void;
+   4 | make<number>(1);
+     |      ~~~~~~
+   5 |       
+---
--- a/internal/rules/no_unnecessary_type_arguments/no_unnecessary_type_arguments.go
+++ b/internal/rules/no_unnecessary_type_arguments/no_unnecessary_type_arguments.go
@@ -105,7 +105,98 @@ var NoUnnecessaryTypeArgumentsRule = rule.Rule{
 			return nil
 		}
 
+		// declarationHasDefaultedTypeParameter reports whether a function-like
+		// declaration has at least one type parameter with a default.
+		declarationHasDefaultedTypeParameter := func(decl *ast.Node) bool {
+			for _, typeParameter := range decl.TypeParameters() {
+				if typeParameter.AsTypeParameterDeclaration().DefaultType != nil {
+					return true
+				}
+			}
+			return false
+		}
+
+		// declarationCannotAcceptArgCount reports whether a function-like
+		// declaration definitely cannot be the overload resolved for a call with
+		// argCount value arguments. It only returns true when certain: the
+		// declaration has fewer parameter slots than the call has arguments and
+		// no rest parameter to absorb the extras. Conservative (returns false)
+		// whenever argCount is unknown (<0) or a rest parameter is present.
+		declarationCannotAcceptArgCount := func(decl *ast.Node, argCount int) bool {
+			if argCount < 0 {
+				return false
+			}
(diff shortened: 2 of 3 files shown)


connorshea:
this is partially redoing #1064. Once that is merged, this will need to be rebased and re-bench'd accordingly to produce an accurate before/after.

camc314:
Closing as this is stale and has conflicts.

PR #1077, opened 2026-07-12
refactor: miscellaneous code cleanup

connorshea:
## Summary

This simplifies some code by using existing helpers and de-duplicating logic in a few places.

Four commits, one per rule, independently revertable:

- **`no-unsafe-argument`** — `argumentCanBeUnsafe` had the same inline spread-assignment closure that the `no-misused-promises` refactor removed; use the existing `ast.IsSpreadAssignment` predicate, spelled as `utils.Some(...)` to match how the sibling `no-unsafe-assignment` rule writes the identical check. Also drops a redundant `Declarations != nil &&` guard (`len` of a nil slice is already 0) and folds a nested `if`.
- **`no-floating-promises`** — `isFunctionParam` re-implemented `hasMatchingSignature`'s walk over `UnionTypeParts` inline with a "has any call signature" test. It now delegates to `hasMatchingSignature` with a named always-true matcher, so the union walk lives in one place.
- **`switch-exhaustiveness-check`** — inline kind-check closure passed to `slices.IndexFunc` → `ast.IsDefaultClause`.
- **`no-floating-promises`** — `isKnownArgumentAt` hand-rolled a contains loop (which the linter flagged); express it as `utils.Some(..., ast.IsSpreadElement)`.

## Validation

Rule test suites pass, and `go build` / `go vet` are clean.

Diagnostics were compared against the VS Code corpus (`src/tsconfig.json`, ~7.3k files, all rules enabled), three runs of `main` vs three runs of this branch, compared as sorted `rule file:line:col` keys: **all six runs produce a byte-identical diagnostic set of 201,348 keys** (single checksum across all six).

### DIFF
--- a/internal/rules/no_floating_promises/no_floating_promises.go
+++ b/internal/rules/no_floating_promises/no_floating_promises.go
@@ -244,12 +244,9 @@ var NoFloatingPromisesRule = rule.Rule{
 			if len(args.Nodes) <= index {
 				return false
 			}
-			for _, arg := range args.Nodes[:index+1] {
-				if ast.IsSpreadElement(arg) {
-					return false
-				}
-			}
-			return true
+			// A spread at or before the index makes the argument at that position
+			// unknowable.
+			return !utils.Some(args.Nodes[:index+1], ast.IsSpreadElement)
 		}
 
 		var isUnhandledPromise func(
--- a/internal/rules/no_floating_promises/no_floating_promises.go
+++ b/internal/rules/no_floating_promises/no_floating_promises.go
@@ -117,18 +117,14 @@ var NoFloatingPromisesRule = rule.Rule{
 			return false
 		}
 
+		anySignature := func(signature *checker.Signature) bool { return true }
+
 		isFunctionParam := func(
 			param *ast.Symbol,
 			node *ast.Node,
 		) bool {
 			t := checker.Checker_getApparentType(ctx.TypeChecker, ctx.TypeChecker.GetTypeOfSymbolAtLocation(param, node))
-
-			for _, part := range utils.UnionTypeParts(t) {
-				if len(utils.GetCallSignatures(ctx.TypeChecker, part)) != 0 {
-					return true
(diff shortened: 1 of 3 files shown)


internal/rules/no_floating_promises/no_floating_promises.go
```diff
@@ -117,18 +117,14 @@ var NoFloatingPromisesRule = rule.Rule{
...
 			node *ast.Node,
 		) bool {
 			t := checker.Checker_getApparentType(ctx.TypeChecker, ctx.TypeChecker.GetTypeOfSymbolAtLocation(param, node))
-
-			for _, part := range utils.UnionTypeParts(t) {
-				if len(utils.GetCallSignatures(ctx.TypeChecker, part)) != 0 {
-					return true
-				}
-			}
-			return false
```
camc314, line 131: I actually don't like this - this change is hiding what the code is actually doing.

internal/rules/no_unsafe_argument/no_unsafe_argument.go
```diff
@@ -63,13 +63,11 @@ func newFunctionSignature(
...
-				restParam = param
-				parameters = parameters[:i]
-				break
-			}
+		if len(param.Declarations) != 0 && utils.IsRestParameterDeclaration(param.Declarations[0]) {
+			// is a rest param
+			restParam = param
+			parameters = parameters[:i]
+			break
 		}
```
camc314, line 73: we probably needs to keep the param.Declarations nil check, as it might be nil if we are given an invalid ast (we don't want to crash tsgolint in this case)

camc314 (changes requested):
two comments - rest looks good!

camc314 (approved):
thanks!

internal/rules/no_floating_promises/no_floating_promises.go
camc314: I reverted it.

PR #1079, opened 2026-07-12
refactor: remove dead code from rules

connorshea:
## Summary

Happy to split this into smaller PRs if preferred, most of the changes are straightforward dead code removal.

Removes code across 11 rules that is unreachable or has no effect. Each item was verified against its call sites before removal — no behavior change, and the existing tests pass unmodified.

Independent of #1078 (no shared hunks); the two can merge in either order.

## What's removed

**prefer-optional-chain** (−90 lines)
- `operandFromNode` had two binary-expression blocks guarded by `isAndChain` and `!isAndChain` whose bodies were token-for-token identical and never read `isAndChain`. Merged them. Because every path in the merged block returns, the `isAndChain` block that followed could never see a binary expression, so it collapsed into the plain-operand return.
- `validateChainRoot` compared the node's operator against the operator its only caller had just read off that same node, and its returned `*ast.BinaryExpression` was discarded. Reduced to the `isInsideJSX` check it actually performs.
- `handleEmptyObjectPattern` re-tested the `||` / `??` operator its only caller already dispatches on.

**no-unnecessary-condition**
- An `else if isPropertyAccess || isElementAccess` nested inside the `else` of an outer chain that had already excluded exactly those kinds — a duplicated branch that could never run.

**consistent-type-exports**
- `checkStarExport` and `analyzeNamedExport` each re-tested the exact negation of the condition the listener dispatches on.

**only-throw-error**
- Checked whether the arrow function it had already proven via `ast.IsArrowFunction` was a spread element. The spread check on the `.then()` `onFulfilled` argument is real and is kept.

**no-deprecated**
- A `break` on the alias flag that the enclosing loop condition already asserts, with no reassignment in between.

**prefer-nullish-coalescing**
- Nil-guards around `isNodeEqual` that duplicate its own `a == nil || b == nil` first line.

**prefer-readonly-parameter-types**
- A `len(properties) > 0` wrapper around nothing but two `range` loops.

**no-unsafe-enum-comparison**
- `len(strings.Split(...)) == 0`, which never holds.

**no-array-delete**, **prefer-includes**
- `if node.Kind != ast.KindX` guards inside listeners keyed on that same kind.

**no-misused-promises**
- An empty `if ast.IsMethodDeclaration(node) {}` body inside a branch where it is already true.

`prefer-includes` also had a `needsParens` switch whose first case re-assigned the zero value. Note this one is *not* a pure deletion: dropping the case list would have made `needsParens` unconditionally true, so the initializer is inverted instead and the kind list is what carries the meaning.

## Test plan

- [x] `go test ./...`
- [x] `go vet ./internal/rules/...`
- [x] `gofmt` clean
### Differential check on real codebases

Built the linter from `main` and from this branch, ran all rules over the corpora below, and compared every diagnostic — rule, file, line/column, exact source range, and the full autofix/suggestion text (not just counts, so a changed fixer would surface). Output is **byte-for-byte identical** on both binaries:

| Corpus | Files | Diagnostics | `main` vs this branch |
| --- | --- | --- | --- |
| `microsoft/vscode` (`src/`) | 7,325 | 201,348 | identical |
| `actualbudget/actual` (`desktop-client`, TSX) | 742 | 16,602 | identical |
| `actualbudget/actual` (`component-library`, TSX) | 427 | 545 | identical |

The vscode count excludes `no-unsafe-type-assertion`, which is nondeterministic across runs independently of this change. Every other rule matched exactly.

The TSX corpora are included because `prefer-optional-chain`'s `validateChainRoot` change turns on `isInsideJSX`, and vscode's `src/` is essentially JSX-free. `prefer-optional-chain` fires 84 times in `desktop-client`, so that path is genuinely exercised.

### DIFF
--- a/internal/rules/consistent_type_exports/consistent_type_exports.go
+++ b/internal/rules/consistent_type_exports/consistent_type_exports.go
@@ -166,9 +166,6 @@ var ConsistentTypeExportsRule = rule.Rule{
 			if exportDecl.IsTypeOnly || exportDecl.ModuleSpecifier == nil {
 				return
 			}
-			if exportDecl.ExportClause != nil && !ast.IsNamespaceExport(exportDecl.ExportClause.AsNode()) {
-				return
-			}
 
 			moduleSymbol := ctx.TypeChecker.GetSymbolAtLocation(exportDecl.ModuleSpecifier.AsNode())
 			if moduleSymbol == nil {
@@ -215,9 +212,6 @@ var ConsistentTypeExportsRule = rule.Rule{
 
 		analyzeNamedExport := func(node *ast.Node) *analyzedNamedExport {
 			exportDecl := node.AsExportDeclaration()
-			if exportDecl.ExportClause == nil || !ast.IsNamedExports(exportDecl.ExportClause.AsNode()) {
-				return nil
-			}
 			if exportDecl.IsTypeOnly {
 				return nil
 			}
--- a/internal/rules/no_array_delete/no_array_delete.go
+++ b/internal/rules/no_array_delete/no_array_delete.go
@@ -48,9 +48,6 @@ var NoArrayDeleteRule = rule.Rule{
 
 		return rule.RuleListeners{
 			ast.KindDeleteExpression: func(node *ast.Node) {
-				if node.Kind != ast.KindDeleteExpression {
-					return
-				}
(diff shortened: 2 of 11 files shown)


camc314 (approved):
thanks!

PR #1088, opened 2026-07-14
fix(utils): resolve relative paths before ComputeConfigFileName to avoid vfs panic

BezSaharaD:
## Summary

Pointing oxlint (with the `typescript` plugin active, so the type-aware `oxlint-tsgolint` backend runs) at a directory outside the project's tsconfig `include` — concretely, `node_modules/<some-package>` via `--no-ignore` — panics inside typescript-go's VFS layer while `TsConfigResolver.FindTsConfigParallel` concurrently walks ancestor directories looking for each file's `tsconfig.json`:

```
panic: vfs: path "node_modules/lucide-react/dist/esm/icons/tsconfig.json" is not absolute [recovered, repanicked]

goroutine 10 [running]:
...
github.com/microsoft/typescript-go/internal/vfs/internal.RootLength(...)
github.com/microsoft/typescript-go/internal/vfs/internal.SplitPath(...)
github.com/microsoft/typescript-go/internal/vfs/internal.(*Common).RootAndPath(...)
github.com/microsoft/typescript-go/internal/vfs/internal.(*Common).FileExists(...)
...
github.com/microsoft/typescript-go/internal/project.(*ConfigFileRegistryBuilder).ComputeConfigFileName(...)
github.com/typescript-eslint/tsgolint/internal/utils.(*TsConfigResolver).work(...)
github.com/typescript-eslint/tsgolint/internal/utils.(*TsConfigResolver).FindTsConfigParallel.func1()
...
Error running tsgolint: "exit status: exit code: 2"
```

This is a hard crash that aborts the **entire** lint run — not a false positive/negative on one file, the whole invocation dies with no results at all, including for files that would otherwise resolve fine.

## Root cause

`TsConfigResolver.FindTsconfigForFile` and `.work()` (called from `FindTsConfigParallel`) pass the caller-supplied file path straight into `ComputeConfigFileName`, which walks ancestor directories (`tspath.GetDirectoryPath` + `ForEachAncestorDirectory`) and hands each candidate directly to `fs.FileExists` — without ever absolutizing the input first. If the caller passes a relative path (e.g. oxlint invoked with a relative CLI target), the whole ancestor walk stays relative, and eventually a relative candidate reaches `internal/vfs/internal.RootLength`/`SplitPath`, which assume an absolute path unconditionally and panic instead of erroring.

Since `FindTsConfigParallel` races multiple goroutines over the file list, which file's resolution hits this first (and thus which directory shows up in the panic message) varies run to run — but the crash itself was 100% reproducible any time the resolver was given a relative path outside its own tsconfig's `include`.

## Fix

Resolve to absolute at the boundary where paths enter this package (`toAbsolute` helper), using `filepath.Join` rather than `tspath.GetNormalizedAbsolutePath` — the latter normalizes to `/` separators, which would change the format of paths this package already returns/compares elsewhere (existing tests build expected paths via `filepath.Join`, which keeps `\` on Windows). `tspath.IsRootedDiskPath` is used to skip already-absolute paths.

## Testing
- Confirmed this isn't just a synthetic edge case: oxlint (1.68.0), which vendors tsgolint as oxlint-tsgolint for its --type-aware/typeAware: true mode, hits this exact panic (identical ComputeConfigFileName → FindTsConfigParallel stack) when linting a real project's own node_modules with type-aware rules on — a common real-world scenario, not an artificial one. The identical files linted with type-aware rules off (bypassing this code path entirely) complete with 0 crashes, isolating the crash to this exact resolver.
- Repro'd the panic against several real `node_modules` packages (lucide-react, axios, zod, clsx) with `--no-ignore` — confirmed 100% reproducible before the fix, 0 panics after, across repeated runs.
- `go test ./internal/...` — all packages pass except:
  - `TestFindTsconfigForFile`, `TestFindTsConfigParallel`, `TestFindTsConfigParallel_UsesAncestorConfigWhenNearestConfigExcludesFile`: confirmed these **already fail identically on an unpatched checkout** on Windows (reverted this commit locally and re-ran) — a pre-existing `/` vs `\` path-separator mismatch coming from `ComputeConfigFileName`'s own internal path handling, unrelated to this change. Not fixed here to keep this PR scoped to the crash.
  - `no_deprecated` rule tests: fail in my environment due to a missing `@types/node` in the test fixtures (I didn't run the full `pnpm install`/e2e setup) — environment gap, not a real failure.

## Disclosure

This fix (root cause investigation, patch, and verification) was developed with Claude's (Anthropic) assistance — I reviewed and tested it locally (crash repro across multiple packages + full `go test ./internal/...` run) before submitting.

### DIFF
--- a/internal/utils/find_tsconfig.go
+++ b/internal/utils/find_tsconfig.go
@@ -33,20 +33,44 @@ func NewTsConfigResolver(fs vfs.FS, currentDirectory string) *TsConfigResolver {
 	}
 }
 
+// toAbsolute resolves fileName against currentDirectory if it isn't already an
+// absolute (rooted) path. Unlike tspath.GetNormalizedAbsolutePath, this keeps
+// the platform's native path separators intact (filepath.Join on Windows keeps
+// `\`) instead of normalizing to `/` — ComputeConfigFileName's ancestor walk
+// only needs an absolute path, not a particular separator style, and callers
+// elsewhere in this package/its tests compare returned config paths against
+// filepath.Join-built strings, so switching separator style would just move
+// the bug from "panics on relative paths" to "returns paths in an unexpected
+// style" for existing consumers.
+func toAbsolute(fileName string, currentDirectory string) string {
+	if tspath.IsRootedDiskPath(fileName) {
+		return fileName
+	}
+	return filepath.Join(currentDirectory, fileName)
+}
+
 // Finds the tsconfig.json that governs the given file
 // Reference: `findOrCreateDefaultConfiguredProjectForOpenScriptInfo` typescript-go/internal/project/projectcollectionbuilder.go:629-671
 func (r *TsConfigResolver) FindTsconfigForFile(filePath string, skipSearchInDirectoryOfFile bool) (configPath string, found bool) {
-	configFileName := r.configFileRegistryBuilder.ComputeConfigFileName(filePath, skipSearchInDirectoryOfFile, nil)
+	// ComputeConfigFileName walks ancestor directories starting from filePath (via
+	// tspath.GetDirectoryPath + ForEachAncestorDirectory) and hands each candidate
+	// directly to fs.FileExists. It does NOT absolutize its input first, so a
+	// caller-supplied relative filePath produces relative candidate paths, which
+	// panic deep in typescript-go's vfs layer (RootLength/SplitPath assume an
+	// absolute path unconditionally). Resolve to absolute here, at the boundary
+	// where paths first enter tsgolint from the caller.
+	absoluteFilePath := toAbsolute(filePath, r.currentDirectory)
+	configFileName := r.configFileRegistryBuilder.ComputeConfigFileName(absoluteFilePath, skipSearchInDirectoryOfFile, nil)
 
 	if configFileName == "" {
 		return "", false
 	}
 
-	normalizedPath := tspath.ToPath(filePath, r.currentDirectory, r.fs.UseCaseSensitiveFileNames())
+	normalizedPath := tspath.ToPath(absoluteFilePath, r.currentDirectory, r.fs.UseCaseSensitiveFileNames())
 
 	// Search through the config and its references
 	// This corresponds to findOrCreateDefaultConfiguredProjectWorker
-	result := r.findConfigWithReferences(filePath, normalizedPath, configFileName, nil, nil)
+	result := r.findConfigWithReferences(absoluteFilePath, normalizedPath, configFileName, nil, nil)
 
 	if result.configFileName != "" {
 		return result.configFileName, true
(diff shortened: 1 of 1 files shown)


camc314 (changes requested):
should we not fix this in oxc? and always send absolute paths when --no-ignore is proided? We can then add an assertion in here that all paths passed to tsgolint were absolute

BezSaharaD:
While we definitely should ensure oxc resolves paths to absolute before passing them down, I believe fixing this at the boundary of tsgolint (this PR) is still highly necessary and robust.
​tsgolint can be used by other tools or standalone, and a library boundary shouldn't hard-panic the entire run when fed a relative path. Handling it gracefully here via toAbsolute prevents catastrophic failures regardless of who the caller is.
​If we only fix it in oxc, any other integration passing relative targets to tsgolint will still hit this exact panic. What do you think about keeping this boundary safety in Go, and optionally adding the absolute-path enforcement in oxc as a companion fix?

camc314:
> ​tsgolint can be used by other tools or standalone, and a library boundary shouldn't hard-panic the entire run when fed a relative path. Handling it gracefully here via toAbsolute prevents catastrophic failures regardless of who the caller is.

tsgolint isn't designed to be used by other tools or standalone, we don't need to account for these use cases.

> ​If we only fix it in oxc, any other integration passing relative targets to tsgolint will still hit this exact panic. What do you think about keeping this boundary safety in Go, and optionally adding the absolute-path enforcement in oxc as a companion fix?

Same as above, we should fix this in oxc, and enforce the invarient that files passed to tsgolint must be absolute.

BezSaharaD:
Understood. I will go ahead and close this PR.
​While pushing invariant enforcement to the caller (oxc) makes tsgolint inherently brittle, I respect your decision to keep the library scope limited. However, designing a Go package that crashes the entire execution run with a raw panic instead of returning a standard error goes against robust Go design principles. Relying on the fragile assumption that no consumer will ever pass a relative path is a questionable choice, but if the maintainers are comfortable accepting catastrophic failures as the default behavior for unexpected inputs, I won't press further.
  
​I'll implement the absolute path enforcement on the oxc side as requested. Thanks for the feedback.

camc314:
> ​While pushing invariant enforcement to the caller (oxc) makes tsgolint inherently brittle, I respect your decision to keep the library scope limited. However, designing a Go package that crashes the entire execution run with a raw panic instead of returning a standard error goes against robust Go design principles. Relying on the fragile assumption that no consumer will ever pass a relative path is a questionable choice, but if the maintainers are comfortable accepting catastrophic failures as the default behavior for unexpected inputs, I won't press further.

It's a fine invariant to have, oxlint is the only confumer of tsgolint.

> ​I'll implement the absolute path enforcement on the oxc side as requested. Thanks for the feedback.

Thanks!

PR #1097, opened 2026-07-19
fix(no-unnecessary-type-assertion): preserve type-changing assertions

### LINKED ISSUE #1094: no-unnecessary-type-assertion: 5 more false positives on 0.25.0 where the autofix breaks tsc (follow-up to #1044)
Identified+reported with the help of AI, reviewed by me.

***

Follow-up to #1044 — 0.25.0 fixed most of the 0.24 regressions for us, but we still hit 5 distinct false-positive patterns where `typescript/no-unnecessary-type-assertion` flags an assertion whose removal makes `tsc` fail. Because the rule's removal is applied with plain `--fix` (no `--fix-dangerously` needed), `oxlint --fix` physically breaks the build.

Impact context: on our monorepo the update rewrote ~100 files; most removals were genuinely redundant casts (nice!), but 7 sites were load-bearing and broke typechecking.

## Environment

- oxlint: 1.74.0
- oxlint-tsgolint: 0.25.0
- typescript: 7.0.2 (same behavior verified against the repro below)
- OS: macOS darwin-arm64

## Setup

`.oxlintrc.json`:

```json
{
  "rules": {
    "typescript/no-unnecessary-type-assertion": "error"
  }
}
```

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "strict": true,
    "noUncheckedIndexedAccess": true,
    "noEmit": true,
    "skipLibCheck": true,
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "bundler",
    "lib": ["esnext"]
  },
  "include": ["*.ts"]
}
```

For each case below: `tsc --noEmit` passes **with** the assertion; `oxlint --type-aware` flags it; after `oxlint --type-aware --fix` removes it, `tsc --noEmit` fails with the quoted error.

## Case 1: `as any` on a property inside an object literal that is itself asserted `as T`

```ts
type AccountId = string & { __brand: 'AccountId' }

interface Req {
  params: Record<string, string>
  query: Record<string, string>
  method: string
}

declare const accountId: AccountId

export const req = {
  params: { id: accountId } as any,
  query: {},
} as Req
```

```
case1.ts:12:10: error typescript(no-unnecessary-type-assertion): This assertion is unnecessary since the receiver accepts the original type of the expression.
```

After autofix:

```
case1.ts(11,20): error TS2352: Conversion of type '{ params: { id: AccountId; }; query: {}; }' to type 'Req' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'method' is missing in type '{ params: { id: AccountId; }; query: {}; }' but required in type 'Req'.
```

The inner `as any` is what makes the outer `as Req` conversion legal (it makes `Req` assignable to the source type). This pattern is very common in test mocks.

## Case 2: widening assertion (`enum → string`) inside `.map()` before `.concat()`

```ts
enum Lang {
  EN = 'en-US',
  SV = 'sv-SE',
}

const LANGUAGE_NAMES: Record<Lang, string> = {
  [Lang.EN]: 'English',
  [Lang.SV]: 'Svenska',
}

export const options = Object.values(Lang)
  .map(lang => ({
    label: LANGUAGE_NAMES[lang],
    value: lang as string,
  }))
  .concat([{ label: 'български език', value: 'bg-BG' }])
```

```
case2.ts:14:11: error typescript(no-unnecessary-type-assertion): This assertion is unnecessary since the receiver accepts the original type of the expression.
```

After autofix:

```
case2.ts(16,39): error TS2769: No overload matches this call.
  The last overload gave the following error.
    Type '"bg-BG"' is not assignable to type 'Lang'.
```

The widening cast deliberately changes the inferred element type of the `.map()` result (`{ value: string }` instead of `{ value: Lang }`) so the subsequent `.concat()` accepts non-enum values.

## Case 3: `unknown` truthiness-narrowed to `{}`, asserted to a record type for destructuring

```ts
type AnyObject = Record<string, any>

declare function isEditArray(item: unknown): item is { type: string }[]

export function getEdits(bucket: { data: unknown }): { type: string }[] {
  const { data } = bucket
  if (!data) return []

  const { edits } = data as AnyObject

  if (!isEditArray(edits)) return []
  return edits
}
```

```
case3.ts:9:21: error typescript(no-unnecessary-type-assertion): This assertion is unnecessary since it does not change the type of the expression.
```

After autofix:

```
case3.ts(9,11): error TS2339: Property 'edits' does not exist on type '{}'.
```

After `if (!data)`, `data` is narrowed from `unknown` to `{}` — the assertion to a record type is required to destructure.

## Case 4: reverse numeric-enum lookup under `noUncheckedIndexedAccess`

```ts
enum AppTracker {
  MOOD = 1,
  ENERGY = 2,
}

export function categoryFor(tracker: string): string {
  return AppTracker[Number(tracker) as AppTracker]
}
```

```
case4.ts:7:21: error typescript(no-unnecessary-type-assertion): This assertion is unnecessary since it does not change the type of the expression.
```

After autofix:

```
case4.ts(7,3): error TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

Indexing the reverse mapping with the enum type yields `string`; indexing with plain `number` under `noUncheckedIndexedAccess` yields `string | undefined` — so the assertion does change the type of the enclosing expression, just not the type of the asserted expression itself.

## Case 5: `as any` on the property that decides which union member an object literal matches

```ts
enum GuideId {
  MEASURE = 'measure',
}

interface Guide {
  id: GuideId
  title: string
  thumbnailUrl: string
}

interface GlossaryItem {
  id: string
  title: string
}

export const item: { content: Guide | GlossaryItem } = {
  content: {
    id: 'how-to-measure' as any,
    title: 'How to measure',
    thumbnailUrl: 'https://example.com/thumb.jpg',
  },
}
```

```
case5.ts:18:8: error typescript(no-unnecessary-type-assertion): This assertion is unnecessary since the receiver accepts the original type of the expression.
```

After autofix:

```
case5.ts(20,5): error TS2353: Object literal may only specify known properties, and 'thumbnailUrl' does not exist in type 'GlossaryItem'.
```

With `id: any` the literal matches `Guide`; with `id: string` it can only match `GlossaryItem`, and excess-property checking then rejects `thumbnailUrl`.

## Notes

- Cases 1, 2, 5 produce the "receiver accepts the original type" message; cases 3, 4 produce "does not change the type" — so the false positives seem to come from (at least) two different code paths.
- Cases 1 and 5 suggest assertions to `any` are risky to flag at all when the asserted expression sits inside an object literal: the assertion changes the *enclosing* literal's inferred/checked type, even when the property itself looks directly assignable. (typescript-eslint skips `as any` entirely in this rule.)

camc314:
## Summary

- preserve assertions whose removal changes enclosing object-literal, union, or generic callback inference
- treat index-signature changes and indexed-access key narrowing under `noUncheckedIndexedAccess` as semantically meaningful
- cover all five unsafe autofix patterns reported in the issue

Fixes #1094

### DIFF
--- a/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
+++ b/internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
@@ -356,11 +356,26 @@ var NoUnnecessaryTypeAssertionRule = rule.Rule{
 			return false
 		}
 
+		isElementAccessArgument := func(node *ast.Node) bool {
+			current := node
+			parent := current.Parent
+			for parent != nil && ast.IsParenthesizedExpression(parent) {
+				current = parent
+				parent = parent.Parent
+			}
+			return parent != nil &&
+				ast.IsElementAccessExpression(parent) &&
+				parent.AsElementAccessExpression().ArgumentExpression == current
+		}
+
 		isTypeUnchanged := func(node *ast.Node, expression *ast.Node, uncast, cast *checker.Type) bool {
 			if uncast == cast {
 				return true
 			}
 			if utils.IsTypeParameter(uncast) && isReceiverOfWriteAccess(node) {
+				return false
+			}
+			if compilerOptions.NoUncheckedIndexedAccess.IsTrue() && isElementAccessArgument(node) {
 				return false
 			}
 
@@ -376,7 +391,7 @@ var NoUnnecessaryTypeAssertionRule = rule.Rule{
 			}
 
 			if (utils.IsTypeFlagSet(uncast, checker.TypeFlagsNonPrimitive) && !utils.IsTypeFlagSet(cast, checker.TypeFlagsNonPrimitive)) ||
-				(hasIndexSignature(uncast) && !hasIndexSignature(cast)) ||
+				(hasIndexSignature(uncast) != hasIndexSignature(cast)) ||
 				containsAny(uncast) ||
 				containsAny(cast) ||
 				(containsTypeVariable(cast) && !containsTypeVariable(uncast)) {
@@ -805,7 +820,9 @@ var NoUnnecessaryTypeAssertionRule = rule.Rule{
 			}
 			objectParent := objectExpr.Parent
 			return objectParent != nil &&
-				(ast.IsSatisfiesExpression(objectParent) ||
+				(ast.IsAsExpression(objectParent) ||
+					ast.IsTypeAssertion(objectParent) ||
+					ast.IsSatisfiesExpression(objectParent) ||
 					(ast.IsCallExpression(objectParent) && objectParent.Parent != nil && ast.IsSatisfiesExpression(objectParent.Parent)))
 		}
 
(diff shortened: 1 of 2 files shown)


camc314:
### Merge activity

* **Jul 19, 9:30 PM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Jul 19, 9:30 PM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Jul 19, 9:38 PM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

PR #1134, opened 2026-08-09
fix(consistent-return): handle never-returning branches

N2H4-xz:
## Summary

- treat calls with a TypeScript `never` result as paths that cannot complete normally
- account for terminating blocks, `if`/`else` branches, and `try`/`catch`/`finally` combinations before reporting an implicit-return mismatch
- add positive and negative regression coverage and update the rule snapshots

## Problem

The type-aware `consistent-return` rule relied on `NodeFlagsHasImplicitReturn` when checking whether a function with value-returning branches could fall through. That syntactic flag does not account for calls whose resolved TypeScript return type is `never`.

As a result, a function that returns a value from a `try` block and terminates from its `catch` block through a call such as `process.exit()` was incorrectly reported as missing a return value.

## Implementation

The exit check now performs a conservative type-aware analysis of the function body before reporting. It recognizes `return`, `throw`, `never`-returning calls, nested blocks, fully terminating `if`/`else` branches, and terminating `try`/`catch`/`finally` combinations.

Unrecognized control-flow constructs continue to use the existing behavior, so the change does not broadly suppress diagnostics.

## Testing

- `go test -count=1 -v ./internal/rules/consistent_return` — 40/40 cases passed
- `just ready` on Ubuntu 24.04 under WSL2
  - formatting passed
  - `golangci-lint run` passed with 0 issues
  - the tsgolint build passed
  - e2e tests passed: 20 passed, 1 skipped
  - `go test ./internal/...` passed in full

## Related issue

https://github.com/oxc-project/oxc/issues/24929

## AI assistance disclosure

OpenAI Codex was used to analyze this issue, implement the patch, add regression tests, and run the validation described above. The submitter selected this specific issue and will be responsible for reviewing the change and responding to maintainer feedback.

### DIFF
--- a/internal/rules/consistent_return/consistent_return_test.go
+++ b/internal/rules/consistent_return/consistent_return_test.go
@@ -176,6 +176,49 @@ func TestConsistentReturnRule(t *testing.T) {
           return;
         }
       `, Options: rule_tester.OptionsFromJSON[ConsistentReturnOptions](`{"treatUndefinedAsUnspecified":true}`)},
+		{Code: `
+        declare function processExit(code?: number): never;
+        function readPackageLock(): unknown {
+          try {
+            return JSON.parse("{}");
+          } catch (error) {
+            console.error(error);
+            processExit(1);
+          }
+        }
+      `},
+		{Code: `
+        declare function abort(): never;
+        function foo(flag: boolean): number {
+          if (flag) { return 42; }
+          abort();
+        }
+      `},
+		{Code: `
+        declare function abort(): never;
+        function foo(flag: boolean): number {
+          if (flag) { return 42; }
+          if (Math.random() > 0.5) { abort(); } else { abort(); }
+        }
+      `},
+		{Code: `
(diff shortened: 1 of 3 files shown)


internal/rules/consistent_return/consistent_return_test.go
```diff
@@ -176,6 +176,49 @@ func TestConsistentReturnRule(t *testing.T) {
...
+        declare function processExit(code?: number): never;
+        function readPackageLock(): unknown {
+          try {
+            return JSON.parse("{}");
+          } catch (error) {
+            console.error(error);
+            processExit(1);
+          }
+        }
+      `},
```
camc314, line 189: an error is reported by typescript-eslint for this test case

camc314 (changes requested):
See my comment on:

https://github.com/oxc-project/oxc/issues/24929#issuecomment-5392306846

camc314:
See issue

PR #1160, opened 2026-08-31
fix(no-unnecessary-boolean-literal-compare): preserve nullable boolean narrowing

sharmila-oai:
When `allowComparingNullableBooleansToTrue` is false, fixing `value === true` can expose `null` or `undefined` where the comparison returned a boolean. Use `!!value` when the result can escape a conditional test, while preserving direct-value simplification inside conditions, including nested `&&` and `||`.

Boolean coercion also preserves TypeScript narrowing. For example, this must still type-check after fixing the comparison:

```ts
declare function acceptsTrue(value: true): boolean;
function check(value: boolean | undefined) {
  return value === true && acceptsTrue(value);
}
```

The fix produces `(!!value) && acceptsTrue(value)`. Using `(value ?? false)` here loses narrowing and produces TS2345. Conditional-context traversal stops at `??` so a fix cannot newly evaluate its fallback.

Related: [typescript-eslint/typescript-eslint#3783](https://github.com/typescript-eslint/typescript-eslint/issues/3783) and its accepted [fix #12365](https://github.com/typescript-eslint/typescript-eslint/pull/12365). Double negation is one of the solutions proposed in the issue and also preserves narrowing.

Validation:

- The existing regression now covers nullable results, narrowing within a conjunction, a nested condition, negation cancellation, and a nullish fallback. It fails with the previous coalescing implementation and passes with coercion.
- Directly applied the native binary's emitted fix and checked original and fixed code with TypeScript 7.0.2: both pass. The coalescing version fails with TS2345.
- `go test ./internal/... -count=1` passes.
- `pnpm --ignore-workspace run test --run` in `e2e`: 20 passed, one existing skip.
- `gofmt` and `git diff --check` pass. Full dprint and golangci-lint checks remain pending CI; the local formatting-plugin download timed out.

### DIFF
--- a/internal/rule_tester/__snapshots__/no-unnecessary-boolean-literal-compare.snap
+++ b/internal/rule_tester/__snapshots__/no-unnecessary-boolean-literal-compare.snap
@@ -242,3 +242,40 @@ Message: This expression unnecessarily compares a boolean value to a boolean ins
 Diagnostic 1: noStrictNullCheck
 Message: This rule requires the `strictNullChecks` compiler option to be turned on to function correctly.
 ---
+
+[TestNoUnnecessaryBooleanLiteralCompareRule/invalid-28 - 1]
+Diagnostic 1: comparingNullableToTrueDirect (3:25 - 3:36)
+Message: This expression unnecessarily compares a nullable boolean value to true instead of using it directly.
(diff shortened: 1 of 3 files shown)


camc314:
@codex review

camc314 (approved):
thanks!

PR #1181, opened 2026-09-14
perf(no-unnecessary-type-assertion): reject incompatible types first

tony-scio:
## Summary

`no-unnecessary-type-assertion` currently searches source and asserted type graphs for nested `any` and type variables before checking whether the two types have the same shape and are mutually assignable.

This changes the order of those predicates:

- reject different properties, type arguments, or assignability first;
- only traverse nested call signatures when the assertion could otherwise be reported;
- add a recursive callable regression case.

The boolean conditions are unchanged. The rule still makes the same decision; it avoids an expensive recursive walk for assertions that clearly change the type.

## Why

Callable types can expose deeply generic parameter and return graphs. A downstream assertion between a generic form-watcher function and a narrower subscription callback spent about 11 seconds recursively looking for `any`, even though mutual assignability rejects the pair immediately.

Moving the recursive checks later reduced that case's measured rule time from a 10.8s median to 0.6ms. The file still has no diagnostic.

## Benchmark

Measured on an Apple Silicon downstream monorepo with 16,572 TypeScript/TSX files, 12 Oxlint threads, and local binaries built from the same `main` commit. Baseline and candidate runs were interleaved. Timings below are Oxlint's reported durations from `--debug=timings`.

| Scope | Baseline runs | Candidate runs | Median change |
| --- | --- | --- | --- |
| Only `no-unnecessary-type-assertion` | 17.3s, 17.5s, 17.1s | 7.9s, 6.7s, 7.0s | **17.3s → 7.0s (-59.5%)** |
| Full lint configuration | 72.7s, 77.4s, 86.8s | 64.6s, 62.9s, 62.8s | **77.4s → 62.9s (-18.7%)** |

The rule's accumulated timing fell from 31.4s to 8.5s in the isolated run and from 28.0s to 3.9s in the full configuration. Accumulated rule timings sum work across threads and can exceed wall time.

No benchmark fixture, profile, or downstream source is included in this PR.

## Validation

- Exact downstream parity: the same 626 rule diagnostics, including locations, labels, fixes, and suggestions.
- Full downstream output: the same 31,362 warnings and 1,103 errors.
- `go test ./internal/rules/no_unnecessary_type_assertion`
- `go test ./internal/...`
- E2E snapshots: 20 passed, 1 skipped.
- `pnpm run fmt`

— sent via Glean Tau


tony-scio:
@auvred

camc314:
@codex review

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -430,12 +427,18 @@ var NoUnnecessaryTypeAssertionRule = rule.Rule{
...
+				!checker.Checker_isTypeAssignableTo(ctx.TypeChecker, uncast, cast) ||
+				!checker.Checker_isTypeAssignableTo(ctx.TypeChecker, cast, uncast) {
 				return false
 			}
 
-			return checker.Checker_isTypeAssignableTo(ctx.TypeChecker, uncast, cast) &&
-				checker.Checker_isTypeAssignableTo(ctx.TypeChecker, cast, uncast)
+			return !containsAny(uncast) &&
+				!containsAny(cast) &&
+				!(containsTypeVariable(cast) && !containsTypeVariable(uncast))
```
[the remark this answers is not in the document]
camc314: I checked this against the parent commit and the current upstream rule, and can confirm the behavior changes. Here's another small repro that doesn't need a custom empty interface:

```ts
function f<T extends () => any>(value: T) {
  const result = value as NonNullable<T>;
}
```

The parent commit and the current typescript-eslint rule both leave this alone. With this PR, it gets an `unnecessaryAssertion` diagnostic and a fix removing the assertion.

`NonNullable<T>` reaches the `T & {}` intersection shortcut. Previously, `containsAny(uncast)` found `any` through the callable constraint and returned before that shortcut. Now the shortcut returns `true` before the moved guards get a chance to run. [Upstream still has those guards before the shortcut](https://github.com/typescript-eslint/typescript-eslint/blob/4a742ff890b7cca8cbf79e78a9a3b345913094c6/packages/eslint-plugin/src/rules/no-unnecessary-type-assertion.ts).

I also tested the `T & Empty<any>` example above: it passes as a valid case on the parent and fails on this PR. Keeping the original guards on the intersection success path, while leaving the reordered checks on the normal path, makes both cases and the existing rule suite pass.

I'd add these as regression cases and preserve that guard before merging. The optimization can still work, but the description's claim that the rule makes the same decision isn't true for the current patch. This establishes a diagnostic/autofix difference; I haven't demonstrated that removing the assertion in these examples breaks the program.

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
```diff
@@ -1103,6 +1103,13 @@ declare function consume(value: unknown): void;
 consume(callable as unknown);
     `},
 		{Code: `
+type RecursiveCallable<T> = (value: T) => RecursiveCallable<{ value: T }>;
+type Subscription = (callback: () => void) => { unsubscribe(): void };
+
+declare const callable: RecursiveCallable<any>;
+const subscription = callable as Subscription;
```
camc314, line 1110: Could we add a fixture that actually exercises the expensive traversal being avoided? This example is useful as a no-diagnostic case, but it takes a cheap exit on both versions:

- Before this change, `containsAny` inspects type arguments before call signatures. `RecursiveCallable<any>` immediately exposes `any`, so it never walks the recursive return type.
- With this change, `haveSameTypeArguments` can reject the pair because `RecursiveCallable<any>` has one type argument and `Subscription` has none. It doesn't reach the mutual-assignability checks either.

That means this test would still pass if the optimization were reverted, and it doesn't reproduce the slow path described in the PR.

A reduced version of the watcher/subscription case would help a lot, ideally with a benchmark comparing the parent and this branch. It needs to get past these immediate exits and make the old nested-type walk do meaningful work. The downstream timings are useful evidence, but without a fixture or profile in the PR we can't reproduce the 10.8s-to-0.6ms result or protect that improvement against later changes. I wouldn't add a wall-clock assertion to the unit test; a separate benchmark would be easier to maintain.

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
```diff
@@ -430,12 +427,18 @@ var NoUnnecessaryTypeAssertionRule = rule.Rule{
 				return false
 			}
 
-			if !hasSameProperties(uncast, cast) || !haveSameTypeArguments(uncast, cast) {
+			// Check shape and assignability before recursively walking nested types. Assertions between
+			// incompatible callable types can otherwise traverse very large generic parameter graphs.
+			if !hasSameProperties(uncast, cast) ||
+				!haveSameTypeArguments(uncast, cast) ||
+				!checker.Checker_isTypeAssignableTo(ctx.TypeChecker, uncast, cast) ||
+				!checker.Checker_isTypeAssignableTo(ctx.TypeChecker, cast, uncast) {
```
camc314, line 435: The ordering makes sense for pairs that assignability rejects quickly. There's also a tradeoff worth covering in the benchmark: these calls now run for pairs that the old `containsAny` or type-variable guards would have rejected first.

Assignability isn't always cheap. For similarly shaped generic or recursive types, it can do a substantial structural comparison, populate relation caches, and hit the checker's depth/complexity limits. So the downstream improvement doesn't by itself tell us whether other shapes get slower.

I haven't reproduced a separate performance regression here, so this is a benchmark suggestion rather than another correctness finding. Alongside the incompatible-callable case, I'd include a same-shape case with an easily found nested `any`, and a recursive pair that requires a deeper assignability comparison. That would give us a better picture of the tradeoff before treating this ordering as a general win.

camc314 (commented):
Hi! Thanks for working on this.

I think the fast rejection is worth keeping, however it looks like there's a couple of regressions caused as a result of changing this. Preserving the guards on that success path should let us keep the optimization without changing those diagnostics.

tony-scio:
@camc314 Thanks, good catches. I added the two regression tests from the thread and fixed the logic:

- `T & Empty<any>`
- `NonNullable<T>` where `T extends () => any`

I also added nested generic/callable `any`, recursive callable, asymmetric type-variable, and direct generic-`any` cases. All added valid cases pass against the parent commit, typescript-eslint 8.70, and the latest version of this PR.

I can add a benchmark, but wasn't sure the most idiomatic way to add in this repo. Can you point me in the right direction?

camc314:
@codex review

camc314:
> I can add a benchmark, but wasn't sure the most idiomatic way to add in this repo. Can you point me in the right direction?

we don't have any benchmarks in here due to the variable nature of the garbage collector which can massively skew results.

You could use go's benchmarks/ or `--debug timings` feature with oxlint and take a look at before/after.


Do you mind replying inline to the rest of the comments - thanks!

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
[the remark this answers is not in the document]
tony-scio: Fixed in `8aaf679`. The intersection branch now evaluates `containsAny(uncast)`, `containsAny(cast)`, and the asymmetric type-variable guard before it can reach the `T & {}` success case.

That keeps both reported assertions unreported and unfixed:

- `T & Empty<any>` returns through `containsAny(cast)`.
- `NonNullable<T>` with `T extends () => any` returns through `containsAny(uncast)`.

I added both as valid regressions, plus nested generic/callable `any` and asymmetric type-variable cases. The normal non-intersection path still performs shape and mutual-assignability rejection before the recursive safety walk.

— sent via Glean Tau

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion_test.go
tony-scio: Replaced the original fixture in `8aaf679` with a finite watcher/subscription case that exercises the intended paths.

Both sides have the same `string` type argument and the same readonly property name, so the fixture passes `haveSameTypeArguments` and `hasSameProperties` and reaches mutual assignability.

I also verified the path with temporary instrumentation: the parent implementation visits 10 nested call signatures during `containsAny` before reaching assignability, while the latest implementation reaches assignability with no preceding recursive safety walk. Both versions produce no diagnostic.

I'll handle the benchmark/tradeoff separately in the other thread.

— sent via Glean Tau

internal/rules/no_unnecessary_type_assertion/no_unnecessary_type_assertion.go
tony-scio: Covered in `f51f8e5` with three table-driven benchmark cases:

- `optimization/incompatible-callables` measures the intended fast rejection.
- `tradeoff/nested-any` uses same-shape generic callables with an easily found nested `any`, so the new ordering performs mutual assignability before the recursive safety check.
- `tradeoff/recursive-assignability` uses recursive generic callable pairs that require structural assignability comparison.

The shared benchmark helper prepares the TypeScript program outside the timer and verifies the expected diagnostic count before measuring repeated single-rule passes.

Against the rebased parent, using 10 interleaved one-second samples with `-cpu=1` and `benchstat`:

- Incompatible callables: 176.83 µs → 38.02 µs, **−78.50%**.
- Nested `any`: 6.287 µs → 6.677 µs, statistically neutral (`p=0.105`); allocations 85 → 84.
- Recursive assignability: 6.369 µs → 6.441 µs, statistically neutral (`p=0.529`); allocations 86 → 85.

So the two counter-shapes exercise the ordering tradeoff without showing a significant slowdown, while the intended case retains a strong improvement.

— sent via Glean Tau

tony-scio:
### Benchmark update

Added a reusable, table-driven Go benchmark in `f51f8e5`:

- `rule_tester.RunRuleBenchmark` prepares the TypeScript program outside the timer, verifies the expected diagnostic count, and measures repeated single-rule lint passes with one worker.
- The rule benchmark covers the intended incompatible-callable optimization, the nested-`any` and recursive-assignability tradeoff shapes raised in review, and common fast/ordinary paths.
- New cases can be added as another named `BenchmarkTestCase`; other rules can reuse the same helper.

Results compare the rebased parent (`dfffb29`) with the PR head (`f51f8e5`) on an Apple M3 Pro using Go 1.26.1. Each side used 10 interleaved one-second samples with `-cpu=1`, analyzed with `benchstat`.

| Case | Parent | PR head | Result |
|---|---:|---:|---:|
| Incompatible callables | 176.83 µs | 38.02 µs | **−78.50%** |
| Nested `any` tradeoff | 6.287 µs | 6.677 µs | Neutral (`p=0.105`) |
| Recursive assignability | 6.369 µs | 6.441 µs | Neutral (`p=0.529`) |
| Same type | 5.772 µs | 5.917 µs | Neutral |
| Direct `any` | 5.106 µs | 5.201 µs | Neutral |
| Property mismatch | 6.171 µs | 5.721 µs | **−7.30%** |
| Type-argument mismatch | 6.311 µs | 5.727 µs | **−9.25%** |
| Structurally equivalent | 6.823 µs | 6.808 µs | Neutral |
| Structurally incompatible | 6.110 µs | 5.675 µs | **−7.12%** |

For the intended incompatible-callable case:

- Bytes: 105.23 KiB → 10.71 KiB, **−89.82%**
- Allocations: 879 → 174, **−80.20%**

Allocations decrease in every benchmark case. Neither tradeoff case shows a statistically significant timing regression.

Validation after rebasing over #1183:

- Build passed
- E2E: 22 passed, 1 skipped
- `go test -count=1 ./...` passed
- `go vet ./...` passed
- The first rebased commit also passes its focused rule tests independently

All review comments have now been replied to inline.

— sent via Glean Tau

camc314 (approved):
thanks you!

PR #1184, opened 2026-09-15
fix(no-unnecessary-type-parameters): make type reference branch terminal

connorshea:
AI Disclosure: The test case here was generated with Claude Code, based on the violation in vscode that was missing before this PR. This PR description was written by me.

The Go port was only returning from the type reference branch when there were type arguments. So any reference without them kept going further into the checks, walking all properties for no reason. The tseslint version of the code stopped at the type reference branch, so we want to match the behavior by moving the `return` outside the length check.

This improves parity on vscode, specifically with [this line](https://github.com/microsoft/vscode/blob/f44f55cde0c1ff6c834bad0bb46397bc48711da8/src/vs/workbench/contrib/browserView/electron-browser/browserEditor.ts#L416) where tseslint was reporting a violation, but tsgolint was not. Now both report.

This ports the fix from #1131 and adds a test case, where `T` is considered unnecessary because it's only used once:

```ts
declare class C {
  m<T>(x: { self: C; arg: T }): void;
}
```

This didn't report in tsgolint, but did for tseslint ([playground](https://typescript-eslint.io/play/#ts=6.0.2&fileType=.tsx&code=CYUwxgNghgTiAEkoGdnwMLwN4Ch7wFsAeAFQD4AKADwC5t5kQIAzO9AbnlgHM6T4AvgEo6ANwD2AS2DscAoA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0ipWsRFCAtonyJoqDJCXQO0SODABfELqA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkAhqqrAO4BSZlemArugDTgakEl2TQ8AOW6MAwgAt00ANYcwXXoIC%2BITUA&tokens=false)). Now tsgolint reports it just like tseslint.

I also had Claude ensure parity of violations before/after on this rule with local checkouts of actual, sentry, canvas-lms, mastodon, ts-toolbelt, bsky/social-app, storybook, material-ui, vite, and vscode. No differences were found other than the one in vscode noted above.

As a nice side-effect, this change also improves the performance of the rule because it's bailing out before doing the extra work.


camc314:
### Merge activity

* **Sep 16, 6:07 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Sep 16, 6:07 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Sep 16, 6:16 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

camc314 (approved):
thanks!

PR #1186, opened 2026-09-16
fix(prefer-optional-chain): preserve narrowing of falsy union members

### LINKED ISSUE #1182: prefer-optional-chain: --fix breaks narrowing when the guarded value's union contains `false` (tsc TS2339)
## Summary

`typescript/prefer-optional-chain`'s autofix rewrites `!x || !x.y` to `!x?.y`. That is correct at runtime, and #1060 notes this shape "fixes correctly". But when `x`'s union contains a **non-nullish falsy** member — most commonly `false`, from a `cond && …` expression — the rewrite is not type-correct: `?.` short-circuits only on `null`/`undefined`, so `false` survives into the property access and the narrowing of `x` itself is lost.

Code that type-checked before the fix fails to compile after it.

typescript-eslint's `prefer-optional-chain` does not report this case.

## Reproduction

`repro.ts`:

```ts
declare const flag: boolean;
declare const node: { querySelector(s: string): Element | null } | null;

export function label(): string | false {
	// `el` is `false | Element | null` -- the `false` comes from `flag &&`
	const el = flag && node?.querySelector('[data-x]');
	if (!el || !el.textContent) return false;
	return el.textContent.trim();
}
```

`tsconfig.json`:

```json
{ "compilerOptions": { "strict": true, "target": "ES2022", "module": "ESNext", "moduleResolution": "bundler", "noEmit": true }, "include": ["repro.ts"] }
```

### Steps

1. `tsc -p .` — clean, no errors.

2. `oxlint --type-aware -A all -D typescript/prefer-optional-chain -- repro.ts`

   ```
   repro.ts:7:6: error typescript(prefer-optional-chain): Prefer using an optional chain expression instead, as it's more concise and easier to read.
   ```

3. Apply `--fix`:

   ```diff
   -	if (!el || !el.textContent) return false;
   +	if (!el?.textContent) return false;
   ```

4. `tsc -p .` — now fails:

   ```
   repro.ts(7,11): error TS2339: Property 'textContent' does not exist on type 'false | Element'.
     Property 'textContent' does not exist on type 'false'.
   repro.ts(8,12): error TS2339: Property 'textContent' does not exist on type 'false | Element'.
     Property 'textContent' does not exist on type 'false'.
   ```

## Expected

No diagnostic for this shape, matching typescript-eslint. `!el || !el.textContent` narrows `el` itself; `!el?.textContent` does not, so the two are not interchangeable once `el` can be `false`.

## typescript-eslint comparison

Running `@typescript-eslint/prefer-optional-chain` 8.67.0 with type information over the same file reports nothing. Adding a second function that *is* safely convertible confirms the rule is active and correctly discriminates:

```ts
declare const plain: { prop?: { deep?: string } } | undefined;

export function g(): string | undefined {
	if (plain && plain.prop && plain.prop.deep) return plain.prop.deep;  // <- only this one is reported
	return undefined;
}
```

tsgolint reports both; typescript-eslint reports only the second.

## Not a duplicate of the earlier reports

- #1060 — outer negation dropped on `x == null || !x.y`, inverting runtime behaviour. That issue explicitly lists `!x || !x.y` as fixing correctly, which is true at runtime; this report is about the type-level consequence for unions containing `false`.
- #829, #585, #615/#616 — different shapes (`Boolean(...)` wrappers, `===` comparisons against literals, general ESLint divergence).

The distinguishing condition here is a union member that is falsy but not nullish, so `?.` cannot stand in for the truthiness check.

## Environment

- oxlint 1.81.0
- TypeScript 7.1.0-dev.20260901.1 (behaviour does not look version-specific; this narrowing is long-standing)
- typescript-eslint 8.67.0 for the comparison

Disclosure, so you can weigh it: the binaries used were locally built from branches rather than releases — a patched oxlint and a fork of tsgolint that adds TypeScript content-mapper support. Neither touches this rule. `internal/rules/prefer_optional_chain` in the build differs from `origin/main` only in the submodule import paths (`microsoft/typescript-go/shim/*` -> `microsoft/TypeScript/tsc/shim/*`), and none of the 26 upstream commits since that branch point modify the rule. I could not rule out an interaction with the newer TypeScript submodule by testing a release build, so please discount accordingly if that seems relevant.

---

Filed by Claude (Claude Code) on @wagenet's behalf, disclosed per the AI usage policy in CONTRIBUTING.md. The prose is Claude's, not his. Every command and output above was executed and copied verbatim rather than reconstructed; the reasoning about `?.` and `false` was checked against `tsc` rather than asserted.

camc314:
For `x: false | { y: string } | null`, rewriting `!x || !x.y` to `!x?.y` introduces a type error because optional chaining does not narrow out `false`.

Following typescript-eslint, reject guards containing `false`, `0`, `0n`, or `''` while collecting operands for `&&` and negated `||` chains. The final operand remains eligible, and safe suffixes can still be converted after an unsafe guard.

Fixes #1182.


camc314:
### Merge activity

* **Sep 16, 6:46 AM UTC**: The merge label '0-merge' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint) once it meets the requirements.
* **Sep 16, 6:49 AM UTC**: `camc314` added this pull request to the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).
* **Sep 16, 6:57 AM UTC**: Merged by the [Graphite merge queue](https://app.graphite.com/merges?org=oxc-project&repo=tsgolint).

